Issue #81, opened 2026-03-09
Agents default to Grep/Read over RNA MCP tools despite guardrail


muness:
## Problem Statement

**Current framing:** "Claude defaults to Grep/Read instead of RNA MCP tools. We need a stronger guardrail."

**Reframed as:** Agents reach for tools they were trained on (Grep, Read) over tools they were told about (search_symbols, graph_query), because training weight beats prompt instructions. The guardrail is a *wish*, not a *constraint* — it describes desired behavior but has no mechanism to enforce or even detect violations.

**The shift:** This isn't a guardrail-wording problem. It's a tool-preference problem baked into the agent's prior. The guardrail says "use RNA tools first" but:
1. The agent's training has thousands of examples of Grep → success
2. The agent has zero training examples of `graph_query` → success
3. Prompt-level instructions compete with training-weight priors, and lose when the agent is "in flow"
4. The violation happened precisely when the agent was deepest in exploratory mode — the moment instructions matter most

### Evidence

When corrected, `graph_query(add_edge, impact, 2)` returned 20 dependents including `build_full_graph`, `load_graph_from_lance`, and the exact edge-production pipeline — in one call. Multiple Grep attempts with `edges\.push` and `add_edge` returned either nothing or noise. **The richer tool was both faster and more informative**, yet the agent didn't reach for it.

The existing guardrail text is clear:
> "Every sub-agent prompt that performs code exploration must include an explicit 'Tool usage requirements' block mandating oh_search_context before file reads, search_symbols for navigation, and LSP for type lookups. Grep/Read are last resort."

The tool descriptions already say "Use this INSTEAD OF Grep/Read." Still wasn't enough.

### Constraints

- **Hard:** Can't retrain Claude's priors. Tool preference from training data is what it is.
- **Hard:** MCP tools aren't part of the model's training — they're runtime-discovered. They'll always start at a disadvantage against built-in tools.
- **Soft:** "Put it in a guardrail" — guardrails are text, not enforcement. A guardrail the agent ignores under cognitive load is decoration.
- **Soft:** Tool descriptions are the main lever — how tools present themselves determines whether the agent reaches for them.

### What this framing enables

- **Mechanism design over policy:** Instead of stronger guardrail wording, build mechanisms that make RNA tools the path of least resistance:
  - Could the system prompt's tool-use instructions be modified to deprioritize Grep/Read when RNA tools are present?
[...]

PR #113, opened 2026-03-10
feat: semantic graph entry points for graph_query

muness:
## Summary

- Adds optional `query` parameter to `graph_query` tool, enabling natural language entry points (e.g. "authentication handler") via vector search against the embedding index
- When `query` is provided and `node_id` is absent, finds top-3 matching code symbols and runs the requested traversal from each
- Fully backward compatible: `node_id` still works exactly as before and takes precedence when both are provided
- Handles gracefully when embedding index is unavailable (suggests using `node_id` instead)

Fixes the core UX friction behind #81: agents can now explore the code graph in a single tool call instead of requiring `search_symbols` + `graph_query`.

## Changes
- `src/server.rs`: `GraphQuery.node_id` changed from `String` to `Option<String>`, new `Option<String>` `query` field added. Handler resolves entry nodes from either source, runs traversal from each, merges/deduplicates results.

## Test plan
- [x] All 194 existing tests pass (`cargo test` — 0 failures)
- [x] Clean build with no errors
- [ ] Manual test: call `graph_query` with `query: "embedding search"` and verify semantic matches appear
- [ ] Manual test: call `graph_query` with `node_id` and verify unchanged behavior
- [ ] Manual test: call with neither parameter and verify error message

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -2327,13 +2327,24 @@ impl rust_mcp_sdk::mcp_server::ServerHandler for RnaHandler {
             "graph_query" => {
                 let args: GraphQuery = parse_args(params.arguments)?;
 
+                // Normalize inputs: treat empty/whitespace strings as None, cap top_k
+                let node_id = args.node_id.as_deref().map(str::trim).filter(|s| !s.is_empty());
+                let query = args.query.as_deref().map(str::trim).filter(|s| !s.is_empty());
+                let top_k = args.top_k.unwrap_or(3).clamp(1, 50) as usize;
+
+                // Reject if both are empty after normalization
+                if node_id.is_none() && query.is_none() {
+                    return Ok(text_result(
+                        "Either node_id or query is required. Provide a stable node ID (from search_symbols) or a natural language query.".to_string()
+                    ));
+                }
+
                 // Resolve entry node IDs: either from explicit node_id or semantic query
-                let (entry_node_ids, entry_header): (Vec<String>, String) = if let Some(ref node_id) = args.node_id {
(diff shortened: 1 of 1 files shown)


muness:
## Adversarial Testing Report

**30 adversarial tests added** (224 total), all passing. Pushed.

### Bugs Found

1. **Empty/whitespace query not guarded** — `query: ""` and `query: "   "` deserialize as `Some`, bypassing the `None` check. Handler embeds an empty string, producing garbage vector search results.
2. **Empty `node_id` takes precedence** — `node_id: ""` is `Some("")`, so the handler uses the empty string path (fails graph lookup) even if a valid `query` was also provided.
3. **`top_k=0` silently returns nothing** — produces "No code symbols matched" with no hint that the parameter was invalid.
4. **`parse_args(None)` unfriendly error** — raw serde error instead of "Either node_id or query is required."
5. **Entry nodes stripped from results lose edge info** — A→B edge invisible when both A and B are entry nodes (confirms review finding).
6. **`top_k * 3` overflow on 32-bit** — `u32::MAX * 3` overflows. Low risk (64-bit only target) but should cap.

### Salvage Recommendation

Input validation belongs at the handler boundary, not scattered through business logic. A `validate_graph_query(&args) -> Result<ValidatedGraphQuery>` that normalizes empty strings to None, caps top_k, and rejects both-empty would prevent 4 of these 6 bugs.

### TODO

- [ ] Guard empty/whitespace query (same pattern as #112 fix)
- [ ] Normalize empty `node_id` to None
- [ ] Reject or default `top_k=0`
- [ ] Cap `top_k` to a reasonable max (e.g., 50)
- [ ] Improve `parse_args(None)` error message

PR #117, opened 2026-03-10
feat: LSP type hierarchy for compiler-accurate implements edges

muness:
## Summary
- Add LSP type hierarchy queries (`textDocument/prepareTypeHierarchy`, `typeHierarchy/supertypes`, `typeHierarchy/subtypes`) to the LSP enrichment pass
- Persist results as `Implements` edges with `ExtractionSource::Lsp` and `Confidence::Confirmed`
- Check `typeHierarchyProvider` capability from raw init JSON (lsp-types 0.97 is missing the `ServerCapabilities` field)
- Integrate into `enrich()` for `Trait`, `Struct`, and `Enum` nodes

## Why
Currently `Implements` edges come from tree-sitter (syntactic `impl Trait for Struct`) and `textDocument/implementation`. For languages with complex type resolution (Rust generics, TypeScript conditional types, Java generics), syntactic edges are incomplete. LSP type hierarchy gives compiler-accurate inheritance and implementation chains, producing edges that tree-sitter alone cannot discover.

## Changes
- `src/extract/lsp.rs`:
  - `LspState.has_type_hierarchy` field to cache capability during init
  - Capability check via `init_result.pointer("/capabilities/typeHierarchyProvider")` (works around lsp-types 0.97 missing the field)
  - Three new helper methods: `prepare_type_hierarchy`, `type_hierarchy_supertypes`, `type_hierarchy_subtypes` (following existing call hierarchy pattern)
  - `enrich_type_hierarchy` orchestrator: prepares items, queries supers/subs, resolves against graph nodes, creates edges
  - `resolve_type_hierarchy_item`: matches LSP items to graph nodes by file+name with position fallback
  - Integration in `enrich()` match arms for `NodeKind::Trait`, `NodeKind::Struct`, `NodeKind::Enum`

## Graceful degradation
- If server lacks `typeHierarchyProvider`, type hierarchy is skipped entirely (no requests made)
- If any request times out or fails, logged at debug level and enrichment continues
- If target node not found in graph, edge is skipped (no dangling edges)
- Existing enrichment behavior unchanged — only adds new edges

## Test plan
- [x] All 194 existing tests pass
- [x] Compiles clean with no warnings
- [ ] Manual test with rust-analyzer on a repo with trait hierarchies (rust-analyzer supports type hierarchy since v0.3.1600)

### DIFF
--- a/docs/lsp-enrichment.md
+++ b/docs/lsp-enrichment.md
@@ -25,3 +25,26 @@ The result: `graph_query(mode: "impact", from: "my_fn")` shows the blast radius 
 | Markdown | marksman | `brew install marksman` |
 
 Plus 31 more: Ruby (solargraph), Java (jdtls), Kotlin, Lua, Zig, Elixir, Haskell, OCaml, Scala, Dart, PHP, Swift, Nix, Terraform, TOML, YAML, and others. Full list in `src/extract/mod.rs`.
+
+## Type Hierarchy Enrichment
+
+When a language server advertises `typeHierarchyProvider`, RNA queries supertypes for each Trait, Struct, and Enum node to create compiler-accurate `Implements` edges (e.g., `MyStruct -> MyTrait`).
+
+**How it works:**
+
+1. During initialization, RNA checks the server's `capabilities.typeHierarchyProvider` field
+2. Enrichment runs as a separate second pass after call hierarchy (Pass 1: calls/implementations/links, Pass 2: type hierarchy batch)
+3. For each eligible node, `prepareTypeHierarchy` resolves the node, then `typeHierarchy/supertypes` discovers parent traits/interfaces
+4. Results are resolved against the graph using name + file + position matching, with tiebreakers for same-named types
+
+**Resilience:**
+
+- A strike counter tracks consecutive failures. After 3 strikes (`MAX_TYPE_HIERARCHY_STRIKES`), type hierarchy is disabled for the rest of the enrichment pass to avoid stalling on broken servers
+- Strikes reset on any successful prepare call
+- Servers that don't support type hierarchy are detected at init and skipped entirely
+
+**Limitations:**
+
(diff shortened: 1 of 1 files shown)


muness:
## TODOs Before/After Merge

### Must fix before merge
- [ ] **Remove dead subtypes branch** — `skip_subtypes` inside `enrich_type_hierarchy` is always true (caller only invokes for Trait/Struct/Enum). The entire subtypes branch (lines 740-775) is unreachable dead code. Remove it; function should only do supertypes.
- [ ] **Extract strike counter helper** — strike-counter logic is copy-pasted in Trait and Struct/Enum match arms. Extract to a helper or closure.
- [ ] **Separate type hierarchy into its own pass** — currently interleaved with call hierarchy in the main `enrich()` loop. Separate pass is cleaner and enables future batching.

### Follow-up (post-merge)
- [ ] **Concurrent LSP requests** — transport uses `&mut self` (sequential). JSON-RPC 2.0 supports multiplexed request IDs. Refactoring transport to `Arc<Mutex<>>` writer + ID-routed response futures would allow true batching. For RNA (~70 requests × <10ms = ~700ms) this is acceptable now, but larger repos (1000+ type nodes) would benefit.
- [ ] **Subtypes for non-Rust languages** — subtypes query skipped for Trait/Struct/Enum (correct for Rust). Other languages (Java interfaces, TypeScript abstract classes) may need subtypes. Re-evaluate when adding more LSP servers.
- [ ] **Redundant dual-gating** — inner `skip_subtypes` check is redundant with caller's match arms. After removing dead code, this resolves itself.

### From merit assessment
- [ ] **Supertypes direction is the real value** — struct→trait edges (what does RustExtractor implement?) upgraded from Detected to Confirmed. ~20 edges for RNA's extractor hierarchy.
- [ ] **Server support verified** — rust-analyzer, clangd, jdtls support type hierarchy. typescript-language-server, pyright, gopls correctly skipped via capability detection.

Issue #25, opened 2026-03-07
Consolidate 20+ tools to ~8 intent-based tools


muness:
## Problem Statement

20+ MCP tools is too many. Agents waste tokens reading descriptions, pick the wrong tool, or fall back to built-in Read/Grep/Bash entirely. Every new capability (LSP, schema, multi-root) adds more tools, compounding the problem.

**The problem isn't that agents can't find tools — it's that tool selection is a tax on every interaction.** The model is a parameter; the tool presentation is the variable.

## Reframing

Instead of "show relevant tools only" (filtering), collapse tools into intent-based interfaces where the server routes internally. The agent describes what it wants; the server picks the mechanism.

## Proposed Consolidation

| Intent tool | Absorbs | How |
|---|---|---|
| `oh_get_context` | `oh_get_outcomes`, `oh_get_signals`, `oh_get_guardrails`, `oh_get_metis` | Already returns all context — remove the 4 individual tools |
| `oh_search_context` | `search_all`, `search_markdown` | Extend to search code/commits too, not just .oh/ artifacts |
| `search_symbols` | `search_code` | Graph-aware superset — delete search_code |
| `graph_query` (new) | `graph_neighbors`, `graph_impact` | Single tool with `mode: "neighbors"/"impact"` param |
| `oh_record` (new) | `oh_record_metis`, `oh_record_signal`, `oh_update_outcome`, `oh_record_guardrail_candidate` | Single tool with `type: "metis"/"signal"/"outcome"/"guardrail"` + appropriate fields |
| `outcome_progress` | — | Keep as-is (unique structural join) |
| `oh_init` | — | Keep as-is (unique scaffolding) |
| `file_history` + `search_commits` | Merge into one | `git_search` with mode param |

**Result: 20+ → 8 tools.** No legacy, no backward compat. Delete the old tools entirely.

## Constraints

- No backward compatibility shims — old tools are deleted from `tools/list` AND from `handle_call_tool_request`
- Each consolidated tool must be strictly better than the sum of what it replaces (not just a wrapper)
- Tool descriptions must be clear enough that agents pick the right one without instructions in AGENTS.md
- `oh_search_context` becoming the universal search entry point means it needs to handle code, markdown, commits, AND .oh/ artifacts

## Acceptance Criteria

- [ ] `tools/list` returns ≤ 10 tools
- [ ] All current functionality remains accessible via the consolidated tools
[...]

PR #128, opened 2026-03-10
feat: show LSP enrichment status in query footers

### LINKED ISSUE #126: Show LSP enrichment status in graph_query / search_symbols footer
## Problem

When graph_query runs before LSP enrichment completes, it returns incomplete results (e.g. "No incoming neighbors" for a trait that has implementors). The agent doesn't know results are stale — it looks like a correct empty answer.

## Evidence

From v0.1.5 functional testing:
- `graph_query(query="Enricher trait", edge_types=["implements"], direction="incoming")` → "No incoming neighbors" (LSP still running)
- Same query 30 seconds later → `Enricher for LspEnricher` (correct)

## Proposed fix

Add LSP enrichment status to the index footer that already appears on search results:

- Current: `*Index: 5050 symbols · last scan just now · schema v3*`
- Proposed: `*Index: 5050 symbols · last scan just now · LSP: pending · schema v3*`

States: `LSP: pending` → `LSP: enriched (1247 edges)` → omit when complete for >30s

This gives agents a signal to retry without blocking the query.

## Files

- `src/server.rs` — footer formatting in graph_query / search_symbols handlers
- `src/extract/lsp.rs` — LSP enricher status (needs a way to expose completion state)

muness:
## Summary

Closes #126

When `graph_query` runs before LSP enrichment completes, agents get empty results that look correct. No signal to retry.

## Solution (from solution-space analysis)

**Option B: Shared `LspEnrichmentStatus` struct** — recommended over simpler AtomicBool (no edge count) or query-time derivation (can't distinguish pending from empty).

- Add `LspEnrichmentStatus` enum (`Pending` / `Enriched { edge_count, completed_at }`) to `RnaHandler`
- Background LSP task updates status on completion
- `format_freshness` renders: `LSP: pending` → `LSP: enriched (N edges)` → omit after 30s
- Two update sites: full build (server.rs ~L1498) and incremental update (server.rs ~L1691)

## Solution-space analysis

See `.oh/sessions/126-lsp-status-solution-space.md`

## Test plan

- [ ] `cargo check` passes
- [ ] `cargo test --lib` passes
- [ ] Query before LSP completes shows `LSP: pending` in footer
- [ ] Query after LSP completes shows `LSP: enriched (N edges)`
- [ ] Footer omits LSP status 30s after enrichment completes

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,17 @@ Agent wrappers for each workflow phase (`oh-aim`, `oh-execute`, `oh-ship`, etc.)
 
 ## Status
 
-**Working today:** 7 MCP tools, 6 CLI subcommands, 22 language extractors, 360+ tests. Structural outcome-to-code joins, LSP-enriched impact analysis with status awareness (query footers show `LSP: pending` during enrichment), cross-language constant search, Metal GPU semantic search with adaptive batch sizing (CPU fallback on Linux), event-driven reindex, persistent index with <1s restarts. Ships as a Claude Code plugin with setup skill and record skill.
+**Working:**
+
+- Structural outcome-to-code joins — agents scope work to declared business outcomes without re-prompting
+- Metal GPU semantic search with adaptive batch sizing — searches code by meaning, not just names; self-tunes to hardware
+- LSP-enriched call/type graph — compiler-grade call hierarchy and type relationships across 37 language servers
+- Query staleness awareness — footers show `LSP: pending` so agents know when results are incomplete
+- Event-driven reindex — background scanner detects changes, re-extracts, re-embeds without blocking queries
+- Persistent index with <1s restarts — LanceDB cache survives process restarts
+- 22 language extractors, cross-language constant search
+- 7 MCP tools, 6 CLI subcommands, 360+ tests
+- Ships as a Claude Code plugin with setup skill and record skill
 
 ### Tested On
 


muness:
## /review findings

**Acceptance criteria (issue #126):**
- [x] `graph_query` now has a freshness footer with LSP status -- the key gap. Both empty-results and results paths include it.
- [x] `search_symbols` and `oh_search_context` also include LSP status via `format_freshness`.
- [x] States are correct: not-started (hidden) -> `LSP: pending` -> `LSP: enriched (N edges)` -> hidden after 30s.
- [x] Two update sites covered: full background build (L1587/L1605/L1659) and incremental update (L1760/L1845).

**Thread safety:**
- `LspEnrichmentStatus` uses `AtomicU8` + `AtomicUsize` + `Mutex<Option<Instant>>`. Wrapped in `Arc` on `RnaHandler`. This is correct and lightweight.
- Memory ordering is correct: `Release` on writes, `Acquire` on reads. The `Mutex` on `completed_at` is only held briefly for reads/writes of a single `Option<Instant>`.
- No contention concern: writers are the background task (infrequent), readers are query handlers (fast lock).

**Code quality:**
- 7 new tests cover: not-started, running, complete, zero edges, format_freshness without/with pending/enriched LSP. Good coverage.
- `format_freshness` signature extended with `Option<&LspEnrichmentStatus>` -- backward compatible, CLI callers pass `None`.
- Clean separation: `LspEnrichmentStatus` is a self-contained struct with no dependencies on the rest of the handler.

**AGENTS.md alignment:**
- No anti-patterns violated. Tool names unchanged. Pluggable extractor pattern preserved (status is orthogonal to enrichers).
- `name-tools-honestly` constraint satisfied: footer describes actual state, not aspirations.

**Minor observations:**
- The `set_complete` method stores edge_count before completed_at before state. This is fine because `footer_segment` reads state first (Acquire), then locks completed_at. A reader that sees COMPLETE will always see the timestamp.
- `unwrap()` on mutex lock in `footer_segment` and `set_complete` -- acceptable since poison only happens on panic, and these mutexes are never held across panicky code.

## /dissent findings

**1. Stuck "LSP: pending" on early return (real bug):**
In the full background build path (L1596), if `bg_graph` is `None` when enrichment starts, the function returns without calling `set_complete`. Status stays stuck at RUNNING = "LSP: pending" forever. This is an edge case (graph should exist by that point since Phase 1 just built it), but it is reachable if Phase 1 fails or is skipped. **Fix:** add `bg_lsp_status.set_complete(0)` before the `return` at L1596.

**2. Could "LSP: pending" cause agent retry spam?**
No. Agents don't have a built-in retry-on-pending loop. The footer is informational -- it tells the agent results may be incomplete, which is better than the status quo (silent incomplete results). An agent *might* choose to retry, but that is correct behavior.

**3. Is 30 seconds the right auto-hide threshold?**
Reasonable. LSP enrichment typically takes 5-15 seconds for medium repos. 30s gives agents a window to see the "enriched" confirmation without it being permanent noise. If anything, it could be shorter (15s), but 30s is conservative and correct.

**4. Mutex contention on query latency?**
No. The `Mutex<Option<Instant>>` holds a single word-sized value. Lock/unlock is sub-microsecond. Query handlers are the only readers, and contention with the single writer (background task) is negligible.

**5. Race condition between set_complete and footer_segment?**
No. The Acquire/Release ordering ensures that if a reader sees `state == COMPLETE`, the `completed_at` write has been committed to memory. The reader then locks the mutex to read the timestamp. This is sound.

**6. Could this break existing tool output parsers?**
Low risk. The footer is inside the existing `*Index: ... · schema vN*` italic line. The LSP segment is inserted as a new `· LSP: ...` segment. Any parser matching the full line literally would break, but the footer was already variable (symbol count, scan age). Agents parse these as informational text, not structured data.

**7. `outcome_progress` does not include LSP status.**
This is acceptable -- `outcome_progress` is a structural join over .oh/ artifacts, not a graph query. LSP enrichment status is irrelevant there.

## Verdict

**PASS WITH CAVEATS**

One item to fix before merge:
- **Stuck pending state:** Add `bg_lsp_status.set_complete(0)` before the early `return` at L1596 in the background build path (where `bg_graph` is `None`). Without this, a failed graph build leaves "LSP: pending" in footers permanently.

Everything else is solid: correct memory ordering, good test coverage, clean separation, addresses the exact problem from issue #126.

muness:
## Step 8: README

Updated — status section reformatted as ranked bullet list. LSP status awareness and adaptive batch sizing mentioned.

## Step 9: Smoke test

Added `run_lsp_status_footer_check()` as smoke check #25. Verifies `format_freshness` renders all 3 LSP states (NotStarted, Pending, Complete) correctly.

## Steps 1-9 complete. Ready for merge.

PR #166, opened 2026-03-13
feat: index JS/TS arrow functions and function expressions as Function nodes

muness:
## Summary
Index JavaScript/TypeScript arrow functions and function expressions assigned to variables as `NodeKind::Function` instead of `NodeKind::Const`, giving agents visibility into the most common function declaration pattern in modern JS/TS.

Closes #158

## Problem
RNA did not index named arrow functions or function expressions assigned to variables. `const handler = (req, res) => { ... }` -- the dominant pattern in modern JS/TS -- was either missed entirely or classified as a constant. `search_symbols("handler")` returned zero results for these definitions.

Arrow functions may represent 60-80% of function definitions in modern JS/TS codebases (React components, Express handlers, Redux reducers, etc.), making the symbol graph actively misleading.

## Solution
**Selected:** Check the value child kind in `lexical_declaration` / `variable_declaration` handlers
**Level:** Local optimum

In both JS and TS extractors, when processing variable declarations:
1. Inspect the `value` child of the `variable_declarator`
2. If the value is an `arrow_function` or `function` expression, emit `NodeKind::Function` with extracted parameter signature
3. Otherwise, preserve existing `NodeKind::Const` behavior for scalar values
4. Also handle class property arrow functions (`public_field_definition` with arrow function value)
5. Wire up cyclomatic complexity scoring for arrow function bodies

## Acceptance Criteria
- [x] `const handler = (req, res) => { ... }` indexed as `NodeKind::Function` named "handler"
- [x] `const transform = function(x) { return x + 1; }` indexed as `NodeKind::Function` named "transform"
- [x] `export const middleware = async (ctx) => { ... }` indexed as Function
- [x] Arrow functions get cyclomatic complexity scoring
- [x] `search_symbols("handler")` finds arrow function definitions
- [x] Class method arrow properties captured
- [x] `let` and `var` declarations with arrow functions also captured
- [x] Existing `NodeKind::Const` behavior preserved for scalar values

## Test Plan
- [x] Unit tests for each arrow function pattern (const, let, var, export, async, class property)
- [x] Tests verify NodeKind::Function classification and correct signature extraction
- [x] Tests verify scalar values still classified as NodeKind::Const
- [x] Both JavaScript and TypeScript extractors covered

## Summary by CodeRabbit

* **New Features**
  * Broader detection of arrow functions and function expressions in JS/TS (let/var/const initializers, exported wrappers, class property arrows), indexed as functions with generated signatures and parent-scope linking.
[...]

PR #174, opened 2026-03-13
fix: emit module Defines edges for imports, consts, and type aliases

### LINKED ISSUE #168: Special-cased nodes (imports, consts, arrow functions) lack module/class Defines edges
## Problem

Nodes created in `collect_js_specials()` and `collect_ts_specials()` are pushed directly to the nodes vector without emitting module-to-node `Defines` edges. This means:

- Import nodes
- Const nodes  
- Arrow function nodes (added in #158 / PR #166)

...all lack the `Defines` edge that connects them to their parent module or class scope. Regular functions extracted by the generic extractor get these edges via `collect_nodes`.

This was flagged by CodeRabbit on PR #166 as a "major" finding, but it's a pre-existing pattern that affects all special-cased nodes, not just arrow functions.

## Impact

Graph queries that traverse `Defines` edges to find "what does this module contain" will miss special-cased nodes. The nodes are still findable via `search_symbols` (name-based search), but not via graph traversal from the module.

## Proposed Fix

After creating nodes in the special handler, also emit a `Defines` edge from the module node to the new node. This requires knowing the module node's `NodeId`, which could be:
1. Passed into `collect_*_specials()` as a parameter
2. Constructed from the file path (module name = file stem)

## Related

- PR #166 (arrow function indexing) -- where CodeRabbit flagged this
- Issue #158 (original arrow function issue)

muness:
## Summary
Add missing module-level `Defines` edges for special-cased nodes (imports, consts, type aliases) so graph traversal from a module node finds all symbols.

Closes #168

## Problem
Nodes created in `collect_js_specials()` and `collect_ts_specials()` were pushed directly to the nodes vector without emitting module-to-node `Defines` edges. Graph queries traversing `Defines` edges to find "what does this module contain" would miss imports, consts, and type aliases. Arrow functions were fixed in PR #166 but the other special-cased node types were not.

## Solution
**Selected:** Add `emit_module_defines_edge()` calls for all remaining special-cased node types
**Level:** local optimum

The `emit_module_defines_edge()` helper already exists from PR #166. This change adds calls for:
- JS/TS scalar `const` declarations (NodeKind::Const)
- JS/TS `import` statements (NodeKind::Import)  
- TS `type` alias declarations (NodeKind::Other("type_alias"))

## Acceptance Criteria
- [x] All special-cased nodes (imports, consts, arrow functions) have Defines edges to their parent module
- [x] Class-scoped special nodes have Defines edges to their parent class (already done in PR #166)
- [x] Graph traversal from module node finds all symbols including special-cased ones
- [x] Existing tests pass

## Test Plan
- [x] `test_const_gets_module_defines_edge` — JS and TS
- [x] `test_import_gets_module_defines_edge` — JS and TS
- [x] `test_type_alias_gets_module_defines_edge` — TS
- [x] `test_all_special_nodes_reachable_from_module` — integration test verifying all node types (JS and TS)
- [x] All pre-existing tests still pass (37 extractor tests, full suite green)

## Summary by CodeRabbit

* **Documentation**
  * Added developer documentation detailing the module-level definition tracking pipeline.

* **Bug Fixes**
  * Improved code extraction accuracy by properly establishing module-level relationships for constants, imports, and type aliases.

* **Tests**
  * Added 7+ tests verifying correct module-level definition tracking and edge relationships across node types.

### DIFF
--- a/src/extract/javascript.rs
+++ b/src/extract/javascript.rs
@@ -964,4 +964,82 @@ function regularFn() {}
             module_defines
         );
     }
+
+    // --- Adversarial tests seeded from dissent (#168) ---
+
+    #[test]
+    fn test_no_duplicate_defines_edges_for_const() {
+        // Dissent: verify no duplicate Defines edges from extractor overlap
+        let extractor = JavaScriptExtractor::new();
+        let code = r#"
+const PORT = 3000;
+"#;
+        let result = extractor.extract(Path::new("src/config.js"), code).unwrap();
+
+        let defines: Vec<_> = result
+            .edges
+            .iter()
+            .filter(|e| {
+                e.kind == EdgeKind::Defines
+                    && e.to.name == "PORT"
+            })
+            .collect();
+        assert_eq!(
+            defines.len(),
+            1,
+            "Should have exactly 1 Defines edge for PORT, not duplicates. Got: {:?}",
+            defines.iter().map(|e| format!("{:?}:{:?} -> {:?}:{:?}", e.from.name, e.from.kind, e.to.name, e.to.kind)).collect::<Vec<_>>()
+        );
+    }
+
+    #[test]
+    fn test_no_duplicate_defines_edges_for_import() {
+        // Dissent: verify no duplicate Defines edges for imports
+        let extractor = JavaScriptExtractor::new();
+        let code = r#"
+import { Router } from 'express';
+"#;
+        let result = extractor.extract(Path::new("src/app.js"), code).unwrap();
+
+        let defines: Vec<_> = result
+            .edges
+            .iter()
+            .filter(|e| {
+                e.kind == EdgeKind::Defines
+                    && e.to.kind == NodeKind::Import
+            })
+            .collect();
+        assert_eq!(
+            defines.len(),
+            1,
+            "Should have exactly 1 Defines edge for import, not duplicates. Got: {:?}",
+            defines.iter().map(|e| format!("{:?}:{:?} -> {:?}:{:?}", e.from.name, e.from.kind, e.to.name, e.to.kind)).collect::<Vec<_>>()
+        );
+    }
+
+    #[test]
+    fn test_exported_const_gets_defines_edge() {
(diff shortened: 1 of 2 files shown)


muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-man
The change is minimal and well-precedented: it adds the same `emit_module_defines_edge()` call pattern that PR #166 already established for arrow functions. The helpers already existed and were tested. This just extends coverage to the remaining special-cased node types.

### Contrary Evidence
1. **Import node names are full statement text** (e.g., `import { Router } from 'express'`), not clean identifiers. The Defines edge `app:Module -> "import { Router } from 'express'":Import` has the full import text as the target name, which is consistent with how imports are already stored but differs from how other node types use clean names.
2. **Potential duplicate Defines edges** -- if a const is also somehow handled by the generic extractor (e.g., future config changes), there could be duplicate edges. However, the generic extractor's `collect_nodes` skips `lexical_declaration` and `import_statement` because they're not in `node_kinds`, so this is not a current risk.
3. **No guard against module node not existing** -- we emit a Defines edge FROM a module node that may not exist as an actual Node in the graph (the module node is synthesized by the generic extractor only if top-level functions/classes exist). However, this is the same pattern used by all existing Defines edges and the graph handles dangling references gracefully.

### Pre-Mortem
1. **Edge duplication if extractors overlap**: If a future change adds `lexical_declaration` to the generic config, consts would get double Defines edges. Low risk -- config changes would need explicit review.
2. **Import node naming inconsistency**: Full import text as node name is established behavior but could confuse graph queries expecting clean names. This is pre-existing, not introduced by this PR.
3. **No negative test for no-edge cases**: We don't test that non-const `let`/`var` scalar declarations (which aren't indexed) DON'T get spurious edges. Low risk since no node = no edge.

### Hidden Assumptions
| Generic extractor won't start handling these node types | Double Defines edges -- but easy to detect in tests |
| Module node synthesized elsewhere in the pipeline | Dangling edge from, but graph handles this |
| `name_str` ownership available after push (clone added) | Compile error would catch this |

Issue #164, opened 2026-03-13
Design pattern detection: naming, structural heuristics, graph matching, topology idioms


muness:
## Problem

Agents using RNA's MCP tools cannot discover architectural intent in a codebase. They can find symbols, edges, and importance scores, but have no way to ask "what patterns organize this code?" The graph already contains the data -- patterns are implicit in the topology but never made explicit. This is an indexing/enrichment problem: close the gap between what RNA's graph knows and what agents can ask about.

## Context

- RNA's Rust extractor already does topology pattern detection for 3 patterns (subprocess, network listener, async boundary) via `detect_topology_patterns` in `src/extract/rust.rs` -- this generalizes that existing capability
- petgraph with typed edges (`Calls`, `Implements`, `DependsOn`, `Defines`, `HasField`, `ReferencedBy`) + PageRank (#145) make graph pattern matching tractable
- Node metadata includes `kind`, `name`, `parent_scope` -- sufficient for both naming and structural heuristics
- No other tool in RNA's competitive space does graph-based design pattern detection

### 1. Naming conventions (LOW effort)

Suffix/prefix matching on node names: `*Factory`, `*Builder`, `*Handler`, `*Adapter`, `*Proxy`, `*Observer`, `*Repository`, `*Strategy`. Store as `metadata["pattern_hint"] = "factory"`.

Surprisingly effective -- developers name things after patterns. Zero false negatives for explicitly-named patterns, and the false positive rate is negligible (if someone names something `FooFactory`, they meant it).

### 2. Structural heuristics (MEDIUM effort)

AST-level checks during extraction:
- **Singleton:** private constructor + static field of own type + public accessor
- **Builder:** methods returning Self/same type (fluent chaining)
- **Factory method:** method returning trait/interface type
- **Observer:** register/subscribe + notify/emit method pairs on same type

### 3. Graph pattern matching (MEDIUM effort)

Subgraph template queries on RNA's existing petgraph. The graph and typed edges already exist -- this adds template matching over them.

- **Factory:** Trait A <- Implements <- multiple types; Function B -> DependsOn -> A (return type), name matches `create/build/make/new_*`
- **Observer:** Struct A has methods matching `register/subscribe` AND `notify/emit`; A -> DependsOn -> Trait B (observer interface)
- **Decorator:** Type A wraps Type B (HasField), both Implement same Trait C
[...]

PR #184, opened 2026-03-13
feat: design pattern hints from naming conventions

muness:
## Summary
Detect design pattern hints from naming conventions and store as `metadata["pattern_hint"]` on graph nodes. This is approach 1 of #164 (naming conventions only).

Relates to #164 (does not close -- other approaches remain as follow-up PRs).

## Problem
Agents using RNA's MCP tools cannot discover architectural intent. The graph contains pattern data implicitly in node names but never makes it explicit.

## Solution
**Selected:** Suffix matching on node names during extraction
**Level:** Band-aid (but surprisingly effective -- developers name things after patterns)

Case-insensitive suffix matching in `collect_nodes` (generic extractor) checks node names against 15 known pattern suffixes: Factory, Builder, Handler, Adapter, Proxy, Observer, Repository, Strategy, Singleton, Decorator, Middleware, Provider, Service, Controller, Manager.

Only applies to node kinds that carry architectural intent: Struct, Trait, Function, Enum, and Other (for Class in Python/TS). Fields, imports, modules, and consts are excluded.

## Acceptance Criteria
- [x] Naming convention detection adds `metadata["pattern_hint"]`
- [x] Matches common GoF and architectural pattern suffixes (15 patterns)
- [x] Visible in search output (metadata is already surfaced by search_symbols/graph_query)
- [x] Tests with known pattern names (14 unit tests including integration test)
- [x] No false positives on unrelated names (only suffix matches, kind-gated)

## Test Plan
- [x] 14 new tests covering: all 15 suffixes, case insensitivity, single-word matches, kind filtering (skips Import/Module/Field/Const), Other kind support, integration with Rust extractor
- [x] Full test suite passes (628 tests)

### DIFF
--- a/src/graph/store.rs
+++ b/src/graph/store.rs
@@ -14,7 +14,7 @@ use arrow_schema::{DataType, Field, Schema};
 /// The server auto-drops and rebuilds all LanceDB tables when this mismatches
 /// the stored version. No manual cache deletion needed.
 /// Also surfaced in the index freshness footer on `search_symbols` and `oh_search_context`.
-pub const SCHEMA_VERSION: u32 = 8;
+pub const SCHEMA_VERSION: u32 = 9;
 
 /// Arrow schema for the `symbols` table.
 ///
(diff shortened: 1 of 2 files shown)


muness:
## Ship Step 4: Adversarial Test
**Result:** All 22 tests pass (14 original + 8 adversarial)

### Adversarial Cases Tested
- **Empty name:** Returns None (no crash)
- **Unicode name:** Non-ASCII characters don't crash `to_ascii_lowercase`; suffix after unicode prefix still matches
- **Overlapping suffix:** "ServiceProvider" correctly matches "provider" (first matching suffix)
- **Substring not suffix:** "FactoryUtil" and "HandlerConfig" correctly return None (pattern word is not a suffix)
- **Snake case:** "my_handler" and "event_observer" correctly match
- **Impl kind excluded:** `NodeKind::Impl` correctly filtered out
- **Cross-language Python:** `class UserRepository` gets hint, `class PlainModel` does not
- **Cross-language TypeScript:** `class AuthMiddleware` gets hint, `class UserDTO` does not

### No Issues Found
The implementation handles all edge cases correctly.

PR #182, opened 2026-03-13
feat: hybrid search — BM25 keyword + vector scoring with RRF

muness:
## Summary

Combines sparse keyword matching (tantivy BM25) with dense vector similarity (MiniLM-L6-v2) using LanceDB's native Reciprocal Rank Fusion, so exact keyword matches rank high while semantic similarity still surfaces conceptually related results.

Closes #165

## Problem

RNA has two disjoint search paths that never combine: sparse name/signature matching and dense vector search. Dense search is bad at exact keywords and negation; sparse search is bad at semantic similarity. Users searching for `EmbeddingIndex` get poor results from vector-only search because the model encodes intent rather than exact tokens.

## Solution

**Selected:** Option A — LanceDB native hybrid search
**Level:** local optimum

Use LanceDB's built-in hybrid search which combines:
1. **FTS index** (tantivy BM25) on `title` + `body` columns — created at index build time
2. **Vector search** (cosine similarity) on embeddings
3. **RRF fusion** — LanceDB internally merges both result sets using Reciprocal Rank Fusion

The hybrid search is the new default for `oh_search_context` and for graph traversal entry-point resolution in `search`. A `search_mode` parameter allows explicit control: `"hybrid"` (default), `"keyword"` (BM25 only), `"semantic"` (vector only).

Falls back gracefully to pure vector search if the FTS index isn't available yet.

## Changes

- `src/embed.rs`: Add `SearchMode` enum, `create_fts_index()` helper, `search_with_mode()` method. FTS index built after `index_all_inner()` and `reindex_nodes()`.
- `src/server.rs`: Add `search_mode` param to `Search` and `OhSearchContext` structs. Wire through `parse_search_mode()` helper. Update tool descriptions.
- `Cargo.toml`: Add `lance-index` direct dependency (was already transitive) for `FullTextSearchQuery`.

## Acceptance Criteria

- [x] `search` tool graph traversal uses hybrid scoring for entry-point resolution
- [x] `oh_search_context` uses hybrid scoring by default
- [x] LanceDB FTS index created alongside vector index at embed time
- [x] API backward compatible (existing `search()` still works, defaults to hybrid)
- [x] Graceful fallback to vector-only if FTS index unavailable
- [x] Tests (SearchMode, parse_search_mode, struct deserialization)

## Test Plan

- [x] `cargo check` passes
- [x] `cargo test` — all 591 tests pass
- [ ] Manual: run server, verify `oh_search_context` returns results
[...]

PR #212, opened 2026-03-14
fix: ensure FTS index on cached embeddings + keyword search smoke test

### LINKED ISSUE #211: fix: ensure FTS index on cached embeddings + keyword search smoke test + node lookup UX
## Problem Statement

When the MCP server starts with no code changes, it reuses the cached embedding table. If that table predates hybrid search (PR #182), it has no FTS indexes. This causes `oh_search_context` with `search_mode: "keyword"` or `"hybrid"` to fail with "FTS keyword search failed".

The smoke test suite never caught this because it only tested vector search, not keyword search.

Additionally, `search(node: "some-id")` without a `mode` parameter returned "Empty query" instead of looking up the node — confusing UX for agent callers.

## Acceptance Criteria

- [x] Cached embedding tables get FTS indexes on startup (idempotent)
- [x] `oh_search_context` with `search_mode: "keyword"` works on cached tables
- [x] Keyword search smoke test exercises `SearchMode::Keyword` in CI
- [x] `search(node: "...")` without mode does batch lookup instead of "Empty query" error

muness:
## Summary

Ensure cached embedding tables get FTS indexes on startup so keyword/hybrid search works on tables predating PR #182. Add keyword search smoke test and fix node lookup UX.

Closes #211

## Problem

When the MCP server starts with no code changes, it reuses the cached embedding table. If that table predates hybrid search (PR #182), it has no FTS indexes. This causes `oh_search_context` with `search_mode: "keyword"` or `"hybrid"` to fail with "FTS keyword search failed". The smoke test never caught this because it only tested vector search. Additionally, `search(node: "...")` without a `mode` parameter returned "Empty query" instead of looking up the node.

## Solution

**Selected:** Targeted fix -- add FTS index creation on cached table reuse + smoke test coverage + node lookup fallback
**Level:** Local optimum

1. **`src/embed.rs`**: New `ensure_fts_index()` public method that creates FTS indexes on an existing embedding table (idempotent via `replace=true`)
2. **`src/server/mod.rs`**: Calls `ensure_fts_index()` when reusing cached embedding table
3. **`src/smoke_test.rs`**: New `keyword_search` smoke test that exercises `SearchMode::Keyword`
4. **`src/server/handlers.rs`**: `search(node: "...")` without mode now does batch lookup instead of "Empty query" error

## Acceptance Criteria

- [x] Cached embedding tables get FTS indexes on startup (idempotent)
- [x] `oh_search_context` with `search_mode: "keyword"` works on cached tables
- [x] Keyword search smoke test exercises `SearchMode::Keyword` in CI
- [x] `search(node: "...")` without mode does batch lookup instead of "Empty query" error

## Test Plan

- [x] All 685 tests pass
- [x] No new clippy warnings in changed files
- [ ] CI green
- [ ] CodeRabbit review clean

## Summary by CodeRabbit

* **New Features**
  * Added single-node batch lookup retrieval for more flexible searches.

* **Improvements**
  * Full-text search indexing is now idempotent and more reliable when reusing indexes.
  * Enhanced keyword search behavior for embedding-based queries.

* **Tests**
  * Added a smoke test path validating keyword embedding search and FTS index readiness.

### DIFF
--- a/src/embed.rs
+++ b/src/embed.rs
@@ -386,7 +386,15 @@ impl EmbeddingIndex {
     pub async fn ensure_fts_index(&self) {
         match self.db.open_table(&self.table_name).execute().await {
             Ok(table) => self.create_fts_index(&table).await,
-            Err(_) => {} // No table to index
+            Err(lancedb::Error::TableNotFound { .. }) => {
+                // No table to index yet.
+            }
+            Err(e) => {
+                tracing::warn!(
+                    "EmbeddingIndex: failed to open table '{}' for FTS ensure: {e:#}",
+                    self.table_name
+                );
+            }
         }
     }
 


muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-man
The implementation is minimal and well-targeted: 40 lines across 4 files solving a real regression path (cached tables missing FTS indexes). Using `replace=true` makes the FTS creation idempotent. The smoke test catches this exact failure mode in CI. The node lookup UX fix is a natural extension of the existing batch retrieval pattern.

### Contrary Evidence
1. **Startup latency:** `ensure_fts_index()` runs on every cold start with a cached table. FTS index creation could add measurable startup time for large tables. However, LanceDB `replace=true` is documented as fast when indexes already exist, and the alternative (failing keyword search) is worse.
2. **Silent error swallowing:** `Err(_) => {}` in `ensure_fts_index` discards the error entirely. If the table exists but can't be opened (permissions, corruption), the failure is invisible. However, the caller (`server/mod.rs`) already has its own error handling for `EmbeddingIndex::new()`, and if `open_table` fails here the table was already verified via `has_table()`.
3. **Smoke test false confidence:** The keyword search test queries "main" -- a term that appears in virtually every repo. It tests FTS index presence but not search quality. However, the goal is regression detection, not quality benchmarking.

### Pre-Mortem
1. **FTS index creation races with concurrent queries:** If an agent sends a keyword search while `ensure_fts_index` is still running, the query could fail. Mitigated by the call order in `server/mod.rs` -- FTS index is created before the index is stored in `embed_index`, so no queries can reach it until FTS is ready.
2. **Table schema mismatch:** If the cached table has a schema from a much older version (missing columns FTS expects), index creation could fail silently. This would be caught by the smoke test on the next run.
3. **Smoke test masks deeper issues:** If `SearchMode::Keyword` returns results but they're wrong/irrelevant, the test still passes. This is acceptable for a regression test.

### Hidden Assumptions
| `has_table()` returning true means `open_table()` will succeed | Low -- both use the same LanceDB connection; failure would indicate corruption |
| `replace=true` is truly idempotent (no performance cost) | Low -- LanceDB docs confirm this; worst case is ~100ms extra on startup |
| `node.unwrap()` is safe after `node.is_some()` check | None -- this is a Rust pattern guarantee |

PR #253, opened 2026-03-15
perf: skip LSP for non-code changes, move incremental enrichment to background

### LINKED ISSUE #250: perf: incremental scan runs LSP enrichment synchronously on non-code file changes
## Problem

When non-code files change (`.oh/` markdown, docs), the incremental scan path in `update_graph_with_scan()` runs LSP enrichment synchronously on the changed nodes. This is wasteful — LSP servers (rust-analyzer, pyright, etc.) produce no edges for markdown files but still spawn, wait, and return nothing.

Additionally, the incremental LSP enrichment at line ~1507 in `src/server/mod.rs` is **synchronous** — it blocks the tool response. The full build path runs LSP in a background task, but the incremental path doesn't.

## What we know

- Full build: LSP runs as background task (non-blocking), produces edges (8461 observed)
- Incremental update: LSP runs synchronously (blocking), wastes time on non-code files
- The "0 edges in Xs" footer was partly caused by a LanceDB persist failure (#251), not just LSP waste — but the waste is real independently

## Acceptance criteria

- [ ] LSP enrichment skips when only non-code files (markdown, `.oh/`) changed in the incremental path
- [ ] Incremental LSP enrichment is non-blocking (runs in background like the full build path)
- [ ] Adding a `.oh/` file and calling `search` returns in <2s

[outcome:context-assembly]

muness:
## Summary
Move incremental LSP enrichment to a background task and scope enrichers to changed languages only, so the graph write lock is released quickly and tool calls are not blocked.

Closes #250

## Problem
When any file changes (including `.oh/` markdown), `get_graph()` acquires the write lock and runs LSP enrichment synchronously inside it. This blocks all tool calls for 90+ seconds, even for markdown-only changes where LSP finishes quickly.

## Solution
**Selected:** Background incremental enrichment + language-scoped enrichers
**Level:** Local optimum

Two key changes:
1. **Background incremental LSP** -- Moved the LSP enrichment block in `update_graph_with_scan()` to a `tokio::spawn` background task, matching the full-build pattern. Write lock is held only for fast in-memory graph mutation, released before slow LanceDB persist and re-embedding.
2. **Language scoping via `enrich_all`** -- The `languages` vec collected from changed nodes is passed to `enrich_all()`, which only invokes enrichers matching those languages. If only `.rs` changed, only rust-analyzer runs. Markdown is NOT skipped -- marksman is a registered enricher that can produce cross-reference edges.

Additionally:
- Fixed pre-existing bug where `upsert_nodes` was captured before PageRank, causing stale importance scores in LanceDB persist.
- Added `EnricherRegistry::supported_languages()` public method and test verifying markdown is in registered enricher languages.

## Acceptance Criteria
- [x] Incremental LSP enrichment is non-blocking (runs in background like full build path)
- [x] Background enrichment holds write lock only for in-memory mutation, not LanceDB persist
- [x] LSP scoped to languages of changed files via `enrich_all`
- [x] Markdown NOT blanket-skipped from LSP (marksman is registered)
- [x] Adding a `.oh/` file and calling `search` returns in <2s (verified: 0.83s)

## Test Plan
- [x] `cargo test` -- 723 tests pass
- [x] `cargo clippy` -- no new warnings
- [x] Manual: install, create `.oh/metis/test.md`, search -- returns in 0.83s

## Summary by CodeRabbit

* **Improvements**
  * Language-scoped enrichment now includes markdown, runs as background tasks, releases write locks sooner, updates node importance after topology changes, re-embeds enriched content as needed, persists deltas outside locks, and snapshots state for background processing.
* **Tests**
  * Added tests verifying multi-language enrichment behavior (markdown and Rust).
* **Public API**
  * Added a method to list all languages supported by registered enrichers.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1526,12 +1526,26 @@ impl RnaHandler {
 
             tokio::spawn(async move {
                 let enricher_registry = EnricherRegistry::with_builtins();
+
+                // Check if any registered enricher supports the delta's languages.
+                // If not, this is expected (e.g. yaml-only change with no yaml LSP) --
+                // don't mark global LSP status as unavailable since other enrichers
+                // (rust-analyzer, marksman) may still be healthy.
+                let supported = enricher_registry.supported_languages();
+                let has_supported_language = languages.iter().any(|l| supported.contains(l));
+
                 let enrichment = enricher_registry
                     .enrich_all(&changed_nodes, &bg_index, &languages, &bg_repo_root)
                     .await;
 
                 if !enrichment.any_enricher_ran {
-                    bg_lsp_status.set_unavailable();
+                    if has_supported_language {
+                        // Enricher was expected to run but didn't -- server likely missing.
+                        bg_lsp_status.set_unavailable();
+                    } else {
+                        // No enricher supports these languages -- not an error.
+                        bg_lsp_status.set_complete(0);
+                    }
                     return;
                 }
 


muness:
## Ship Step 1: Review
**Verdict:** CONTINUE

### Acceptance Criteria
- [x] LSP enrichment skips when only non-code files changed in the incremental path -- `is_lsp_worthy_language()` filter at line 1499 excludes markdown/yaml/json/toml/text/unknown
- [x] Incremental LSP enrichment is non-blocking (runs in background like full build path) -- moved to `tokio::spawn` at line 1515
- [ ] Adding a `.oh/` file and calling `search` returns in <2s -- needs manual verification
- [x] Background enrichment holds write lock only for in-memory mutation, not LanceDB persist -- lock dropped before `persist_graph_incremental` and `reindex_nodes`

### Alignment Check
- **Still necessary:** Yes. The synchronous LSP enrichment inside the write lock is the primary cause of 90s+ tool call stalls.
- **Still aligned:** Yes. Targets [outcome:context-assembly] -- faster context delivery.
- **Mechanism clear:** Three orthogonal fixes: (1) language filter skips non-code, (2) background task for incremental LSP, (3) full-build language filter.
- **Changes complete:** Implementation covers all three fixes. Unit test added for `is_lsp_worthy_language`.
- **No drift detected.**

### Guardrail Check
- `extract-fully-at-parse-time`: N/A -- this PR changes enrichment scheduling, not extraction
- `no-parallel-cargo-agents`: N/A -- no parallel builds in this PR

muness:
## Ship Step 4: Adversarial Test

**Seeded from dissent finding:** "Non-code languages produce zero useful LSP edges" assumption, and case sensitivity edge cases.

### Tests added:
1. **`test_is_lsp_worthy_language_covers_all_registered_lsp_servers`** -- verifies every code language in `EnricherRegistry::with_builtins()` (34 languages across tiers 1-3) is correctly classified as LSP-worthy. Ensures no code language is accidentally filtered out.

2. **`test_is_lsp_worthy_language_edge_cases`** -- tests case sensitivity and whitespace handling. The denylist uses exact lowercase match, so `"Rust"`, `"MARKDOWN"`, `" markdown"`, `"markdown "` all pass through (not denied). This is correct because language strings are normalized to lowercase by extractors.

**Result:** All 3 `is_lsp_worthy_language` tests pass. The denylist approach (blocking known non-code languages rather than allowlisting code languages) is conservative -- new languages are LSP-worthy by default, which is the safe direction.

muness:
## Ship Step 7a: Manual Verification

**Full test suite:** 725 tests pass (723 existing + 2 new adversarial tests).

**Code path verification:**
- `is_lsp_worthy_language` correctly filters markdown/yaml/json/toml/text/unknown
- The `tokio::spawn` block in `update_graph_with_scan` mirrors the existing background enrichment pattern from `build_full_graph` (lines 931-1048)
- Write lock acquisition/release pattern verified: lock acquired at bg_graph.write().await, held only for in-memory mutations, dropped before persist_graph_incremental
- PageRank recomputed in both the main path (line 1605) and the background task (after LSP edges added)

**Manual test not possible in CI context:** The `<2s search after .oh/ file creation` acceptance criterion requires a running MCP server. This must be verified post-merge by the developer.

muness:
## Ship Step 7b: Delivery Verification

**N/A** -- This PR does not add agent-visible data. It changes the scheduling/timing of LSP enrichment (synchronous -> background) and adds a language filter. No new metadata, no new Arrow columns, no new MCP tool output fields.

muness:
## Additional: scope LSP enrichment to changed languages only

The incremental enrichment should only run LSP enrichers for the languages of the files that actually changed. If only `.rs` files changed, only run rust-analyzer — don't spawn pyright, tsserver, gopls, marksman.

The current code at line ~1497 already collects `languages` from changed nodes, but `enrich_all` iterates ALL registered enrichers regardless. The language filter should be passed through to only invoke matching enrichers.

muness:
@coderabbitai review

muness:
## Ship Step 3: Fix (CodeRabbit round 2)

### Fixed: "Don't mark global LSP status unavailable for unsupported deltas" (line 1535)
**Fixed in 724291a.** When only unsupported languages change (e.g. yaml with no yaml LSP), we now call `set_complete(0)` instead of `set_unavailable()`. Only mark unavailable when a supported language's enricher fails to run. Uses the new `supported_languages()` method to check the intersection.

### Deferred: "Outer write guard held across slow I/O" (line 1527)
**Pre-existing, out of scope.** The outer `get_graph()` write guard (line 182) wraps the entire `update_graph_with_scan()` call, including extraction persist and re-embedding. This PR moved the LSP enrichment (the 90s+ bottleneck) to a background task. The remaining persist + re-embed is seconds, not minutes. Restructuring `get_graph()` to release the outer lock after in-memory mutations would be a larger refactor -- tracked as a future optimization.

### Deferred: "Reject stale enrichment results" (line 1603)
**Architectural concern, out of scope.** If a file is edited while the background LSP task is running, the older task could reinsert stale edges. This is a valid concern but: (1) the same pattern exists in the full-build background path since initial implementation, (2) `persist_graph_incremental` uses merge_insert (upsert), so stale writes are overwritten by the next scan, (3) adding generation/epoch tracking requires cross-cutting changes to the enrichment API. Worth tracking as a separate issue.

PR #256, opened 2026-03-15
fix: LSP status reflects persist failures + foreground propagation

### LINKED ISSUE #235: fix: persist failures silently ignored across all persist paths
## Problem

Two related failure modes around persist reliability:

### 1. Persist failures silently ignored (original issue)

CodeRabbit flagged on PR #233: persist failures (writing to LanceDB) are silently ignored across all persist paths. When `incremental_persist` or similar LanceDB write operations fail, the error is logged but execution continues as if nothing happened. This means:
- The graph reports "enriched (N edges)" but the edges may not have been written to disk
- Next restart loads stale data with no indication that the last persist failed
- Silent data loss

### 2. Scanner persists state before graph update succeeds (NEW — bigger issue)

`Scanner::scan()` persists the updated scan state (file mtimes, content hashes) to `.oh/.cache/scan-state.json` at the end of the scan call (scanner.rs ~line 521), **before** the caller (`update_graph_with_scan`) processes the results. If the graph update fails or takes a different code path:

- The scan state records "I've seen these files at these mtimes"
- But the graph/index never processes them
- The next scan sees no changes (mtimes match) — **the changes are permanently lost**
- The index stays stale with no error, no warning, no recovery path

This was observed in practice: `.oh/` files were created/modified, the scan state tracked them, but `oh_search_context` returned no results and the freshness footer stayed frozen at the initial scan time.

**Root cause:** The scanner owns persistence of its own state, but the caller owns whether changes were successfully processed. These two concerns are coupled in the wrong order.

## Fix approach

1. **Defer scan state persistence:** `scan()` should return results without persisting state. The caller persists state only after successful processing. Alternatively, use a two-phase approach: `scan()` returns a `ScanResult` with a `commit()` method that persists state.

2. **Propagate persist errors:** Persist failures should propagate as errors, not just warnings. The graph should not report completion if persistence failed.

## Acceptance criteria
- [x] Scanner state is persisted only after the graph update succeeds
- [x] If graph update fails, next scan re-detects the same changes (no silent loss)
- [x] Persist failures propagate as errors (not just warnings)
- [ ] `scan --full` reports persist failures visibly
- [x] MCP server logs persist failures at ERROR level
- [ ] LSP completion status reflects whether edges were actually persisted
- [x] Freshness footer updates when incremental scan processes changes

## Source

- CodeRabbit review on PR #233 (original persist failure)
- Discovered during live `.oh/` corpus curation session — new guardrail/outcome files invisible to `oh_search_context` after creation

[outcome:agent-alignment]

muness:
## Summary

LSP enrichment persist failures are no longer silently ignored. The footer now shows "persist failed" when LanceDB writes fail, and the foreground pipeline propagates the error.

Closes #235

## Problem

When `persist_graph_incremental` failed after LSP enrichment, the status was set to `set_complete(edge_count)` regardless. The footer showed "enriched (N edges)" even though those edges may not have been written to disk. In the foreground `--full` pipeline, the error was logged but execution continued as if nothing happened.

## Solution

**Selected:** Add persist-failure awareness to `LspEnrichmentStatus`
**Level:** local optimum

1. New `persist_failed` flag on `LspEnrichmentStatus` + `set_complete_persist_failed()` method
2. Footer renders "enriched (N edges, persist failed)" when persist fails (never auto-hidden)
3. All three background LSP enrichment paths use `set_complete_persist_failed()` on error
4. Foreground pipeline returns the error so `scan --full` reports it visibly

## Acceptance Criteria

- [x] LSP completion status reflects whether edges were actually persisted
- [x] `scan --full` reports persist failures visibly (propagates error)
- [x] Background LSP enrichment paths distinguish "enriched" from "persist failed"
- [x] 726 tests pass, no new clippy warnings

## Test Plan

- [x] `cargo test` -- 726 passed, 0 failed
- [x] `cargo clippy` -- no new warnings
- [x] New unit tests: `test_lsp_status_persist_failed_shows_in_footer`, `test_lsp_status_persist_failed_no_auto_hide`, `test_lsp_status_persist_failed_cleared_on_success`

[outcome:agent-alignment]

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified and more explicit handling of persistence outcomes across background and foreground enrichment paths, ensuring a full re-persist after schema migration and surfacing failures to status.
  * Added a persistent "persist failed" indicator in the status footer which shows edge counts and prevents automatic dismissal when set.

* **Tests**
  * Added unit tests to verify persistence-failure tracking, footer display, and auto-hide behavior.

### DIFF
--- a/src/server/enrichment.rs
+++ b/src/server/enrichment.rs
@@ -476,7 +476,24 @@ impl RnaHandler {
                     match persist_graph_incremental(
                         &lsp_repo_root, &upsert_nodes, &persist_edges, &[], &[],
                     ).await {
-                        Ok(_) => bg_lsp_status.set_complete(edge_count),
+                        Ok(true) => {
+                            tracing::info!("[background] LSP enrichment: schema migrated; performing full persist");
+                            let snapshot = {
+                                let g = bg_graph.read().await;
+                                g.as_ref().map(|gs| (gs.nodes.clone(), gs.edges.clone()))
+                            };
+                            if let Some((nodes, edges)) = snapshot {
+                                if let Err(e) = persist_graph_to_lance(&lsp_repo_root, &nodes, &edges).await {
+                                    tracing::error!("[background] LSP enrichment: full persist after migration failed: {}", e);
+                                    bg_lsp_status.set_complete_persist_failed(edge_count);
+                                } else {
+                                    bg_lsp_status.set_complete(edge_count);
+                                }
+                            } else {
+                                bg_lsp_status.set_complete(edge_count);
+                            }
+                        }
+                        Ok(false) => bg_lsp_status.set_complete(edge_count),
                         Err(e) => {
                             tracing::error!("Background LSP enrichment persist failed: {}", e);
                             bg_lsp_status.set_complete_persist_failed(edge_count);
@@ -586,7 +603,24 @@ impl RnaHandler {
                 match persist_graph_incremental(
                     &bg_repo_root, &upsert_nodes, &persist_edges, &[], &[],
                 ).await {
-                    Ok(_) => bg_lsp_status.set_complete(edge_count),
+                    Ok(true) => {
+                        tracing::info!("[background] LSP enrichment: schema migrated; performing full persist");
+                        let snapshot = {
+                            let g = bg_graph.read().await;
+                            g.as_ref().map(|gs| (gs.nodes.clone(), gs.edges.clone()))
+                        };
+                        if let Some((nodes, edges)) = snapshot {
+                            if let Err(e) = persist_graph_to_lance(&bg_repo_root, &nodes, &edges).await {
+                                tracing::error!("[background] LSP enrichment: full persist after migration failed: {}", e);
+                                bg_lsp_status.set_complete_persist_failed(edge_count);
+                            } else {
+                                bg_lsp_status.set_complete(edge_count);
+                            }
+                        } else {
+                            bg_lsp_status.set_complete(edge_count);
+                        }
+                    }
+                    Ok(false) => bg_lsp_status.set_complete(edge_count),
                     Err(e) => {
                         tracing::error!("Background LSP enrichment persist failed: {}", e);
                         bg_lsp_status.set_complete_persist_failed(edge_count);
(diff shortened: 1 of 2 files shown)


muness:
## Step 1: Review

**Aim:** Stop silently ignoring persist failures in LSP enrichment paths.

### Acceptance Criteria Check (from #235)

| Criterion | Status | Notes |
|-----------|--------|-------|
| Scanner state persisted only after graph update succeeds | Already done (PR #238) | Not in scope |
| If graph update fails, next scan re-detects changes | Already done (PR #238) | Not in scope |
| Persist failures propagate as errors (not warnings) | Done | All paths now use `tracing::error!`; foreground path returns `Err` |
| `scan --full` reports persist failures visibly | Done | Foreground pipeline propagates error via `return Err(e.context(...))` |
| MCP server logs persist failures at ERROR level | Done | Already `tracing::error!` in all paths |
| LSP completion status reflects whether edges were actually persisted | Done | New `set_complete_persist_failed()` method; footer shows "persist failed" |
| Freshness footer updates when incremental scan processes changes | Already done (PR #238) | Not in scope |

### Drift Check
- **Scope:** Tight -- only touches `state.rs` and `enrichment.rs`, no unnecessary changes
- **Solution:** Matches the selected approach (persist_failed flag on LspEnrichmentStatus)
- **Goal:** Aligned with #235 remaining items

### Verdict: **Continue**

All remaining acceptance criteria from #235 are addressed. Implementation is minimal and correct.

muness:
## Step 2: Dissent

**Devil's advocate examination of this change:**

1. **Could `persist_failed` flag become stale?** The flag is cleared on successful `set_complete()`. If a background enrichment fails, then a later incremental enrichment succeeds, the flag is properly cleared. The ordering is correct because `set_complete` stores `false` to `persist_failed` before transitioning state.

2. **Is the foreground error propagation too aggressive?** When `scan --full` hits a persist failure after LSP enrichment, it returns an error. The graph build and embedding have already succeeded and been persisted. The LSP edges are in-memory but not on disk. Returning an error is the right signal -- it tells the user that the scan did not fully succeed. The graph is still queryable (in-memory) until restart.

3. **Race condition in `persist_failed` flag?** The flag uses `AtomicBool` with `Release`/`Acquire` ordering, matching the pattern of `edge_count`. Since `set_complete` and `set_complete_persist_failed` are always called from a single code path (not concurrently for the same enrichment pass), there's no data race.

4. **Should background persist failures be retried?** Not in this change -- retry logic would add complexity. The flag makes the failure visible, which is the acceptance criterion. Retry can be a follow-up.

**Verdict:** No blocking concerns. Proceed.

muness:
## Step 4: Adversarial Test

**Test cases verified:**

1. **`test_lsp_status_persist_failed_shows_in_footer`** -- Confirms footer includes both edge count and "persist failed" text.

2. **`test_lsp_status_persist_failed_no_auto_hide`** -- Confirms persist failure footers are never auto-hidden (unlike normal complete footers which fade after 30s). This ensures agents always see the failure.

3. **`test_lsp_status_persist_failed_cleared_on_success`** -- Confirms that a subsequent successful `set_complete()` clears the `persist_failed` flag. This prevents stale failure indicators after recovery.

4. **Existing concurrent state tests** (`test_adversarial_concurrent_state_transitions`) -- The new `AtomicBool` field follows the same memory ordering pattern as existing atomic fields. Concurrent access is safe.

5. **Foreground error propagation** -- The `run_pipeline_foreground` path returns `Err` with context when persist fails. This causes `scan --full` CLI to report the failure instead of silently continuing.

All 726 tests pass. No adversarial issues found.

muness:
## Step 3: Fix (CodeRabbit feedback)

Addressed both CodeRabbit review findings:

1. **Handle `Ok(true)` schema migration in all LSP persist paths** -- All four LSP enrichment persist sites now match the existing pattern from the background scanner: branch on `Ok(true)` to perform a full persist via `persist_graph_to_lance()`, mark `set_complete_persist_failed()` if that full persist fails.

2. **Improve `test_lsp_status_persist_failed_no_auto_hide`** -- Test now backdates `completed_at` past the 30s auto-hide window to properly verify the no-auto-hide behavior for persist failures.

726 tests pass. Pushed as commit de00396.

PR #259, opened 2026-03-15
fix: apply search_mode to flat code symbol search

### LINKED ISSUE #219: search_mode silently ignored on flat search
## Problem Statement

When agents call `search(query: "auth handler", search_mode: "semantic")`, the `search_mode` parameter is silently ignored for flat search (no `mode` parameter). The tool accepts the parameter without error but always uses name/signature string matching, regardless of the requested search mode. Agents get name-matched results with no indication that their ranking preference was ignored.

The tool description explicitly states: "Only affects how entry nodes are found when using query + mode; flat search always uses name/signature matching." But agents naturally expect `search_mode` to apply to all searches. This is friction F1 from the v0.1.7 exercise.

## Acceptance Criteria

- [x] `search(query: "auth handler")` uses hybrid ranking (BM25 + vector RRF) by default — matching the graph traversal default
- [x] `search(query: "auth handler", search_mode: "keyword")` uses BM25-only ranking
- [x] `search(query: "auth handler", search_mode: "semantic")` uses vector-only ranking
- [x] Flat search falls back to name/signature matching gracefully if the embedding index is not ready
- [x] The `search` tool description no longer says flat search always uses name/signature matching
- [x] Existing sort_by (complexity, importance) and filter (kind, file, language, synthetic, min_complexity) parameters continue to work with embed-ranked results
- [x] Tests cover the new ranking paths

muness:
## Summary
Flat search now uses the embedding index for code symbol ranking, respecting `search_mode` (hybrid/keyword/semantic). Falls back to name/signature matching when the embedding index is not ready.

Closes #219

## Problem
When agents call `search(query: "auth handler", search_mode: "semantic")`, the `search_mode` parameter was silently ignored for flat search (no `mode` parameter). The tool accepted the parameter without error but always used name/signature string matching, regardless of the requested search mode.

## Solution
**Selected:** Use embedding index for flat code symbol search when available
**Level:** Local optimum -- extends existing embed search infrastructure to flat search path

- Extract code symbol search into `flat_code_symbol_search()` helper
- When embed index is available and query is non-empty, use `search_with_mode` to find code symbols ranked by the embedding index
- Filter embed results to `code:*` kinds, resolve back to graph nodes
- Apply post-filters (kind, language, file, root, synthetic, min_complexity)
- Fall back to name/signature matching when embed index is unavailable or not ready
- Preserve embed ranking order (skip name-match re-ranking for embed results)
- sort_by overrides (complexity, importance) still apply on top
- Updated tool description to reflect that search_mode applies to all search paths

## Acceptance Criteria
- [x] `search(query: "auth handler")` uses hybrid ranking (BM25 + vector RRF) by default
- [x] `search(query: "auth handler", search_mode: "keyword")` uses BM25-only ranking
- [x] `search(query: "auth handler", search_mode: "semantic")` uses vector-only ranking
- [x] Flat search falls back to name/signature matching gracefully if the embedding index is not ready
- [x] The `search` tool description no longer says flat search always uses name/signature matching
- [x] Existing sort_by (complexity, importance) and filter (kind, file, language, synthetic, min_complexity) parameters continue to work with embed-ranked results
- [x] Tests cover the new ranking paths (11 new tests)

## Test Plan
- [x] `cargo test` -- 741 tests pass (11 new)
- [x] `cargo clippy` -- no errors
- [x] New tests cover: fallback name matching, signature matching, kind filter, language filter, file filter, sort_by complexity, sort_by importance, search_mode parsing, empty query handling, min_complexity filter, search_mode with no embed

[outcome:context-assembly]

## Summary by CodeRabbit

* **Documentation**
  * Updated search tool documentation to reflect embedding-based ranking (hybrid BM25+vector) with fallback to name/signature matching
  * Clarified search_mode configuration applies across all search paths

* **New Features**
  * Implemented embedding-based semantic search with automatic fallback behavior when index is unavailable

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ The system compounds from here. Agents use `search` to discover relevant context
 
 | Tool | What it's for |
 |------|--------------|
-| `search` | All-in-one search: code symbols by name/signature, .oh/ artifacts and commits via embedding index, markdown sections, and graph traversal with `mode` (neighbors, impact, reachable, tests_for). Supports `include_artifacts` and `include_markdown` (both default true), `artifact_types` filter, `compact: true` (~25x fewer tokens), batch retrieval via `nodes: [...]`, sorting by relevance/complexity/importance, and `search_mode` (hybrid/keyword/semantic). |
+| `search` | All-in-one search: code symbols ranked by embedding index (hybrid BM25+vector by default, falls back to name/signature matching while index builds), .oh/ artifacts and commits, markdown sections, and graph traversal with `mode` (neighbors, impact, reachable, tests_for). Supports `include_artifacts` and `include_markdown` (both default true), `artifact_types` filter, `compact: true` (~25x fewer tokens), batch retrieval via `nodes: [...]`, sorting by relevance/complexity/importance, and `search_mode` (hybrid/keyword/semantic) applies to all search paths. |
 | `repo_map` | Repository orientation: top symbols by PageRank importance, hotspot files, active outcomes, entry points. One call replaces an exploratory loop. |
 | `outcome_progress` | Connect business outcomes to code: outcome → tagged commits → changed files → symbols. Optional `include_impact: true` for risk-classified blast radius. |
 | `list_roots` | Show which workspace roots are configured and their scan status |
@@ -210,7 +210,7 @@ search(node="<id>", mode="impact")
 
 | Command | What it does |
 |---------|-------------|
-| `search <query>` | Search symbols by name/signature, filter by kind/language/file |
+| `search <query>` | Search symbols via embedding index (hybrid/keyword/semantic), filter by kind/language/file |
 | `graph --node <id> --mode <mode>` | Traverse neighbors, impact analysis, or reachability |
 | `scan --repo <dir>` | Scan + extract + embed + persist |
 | `scan --repo <dir> --full` | Full pipeline including LSP enrichment, with visible phase-by-phase output |


muness:
## Ship Step 1: Review
**Verdict:** CONTINUE

### Acceptance Criteria
- [x] `search(query: "auth handler")` uses hybrid ranking (BM25 + vector RRF) by default -- `parse_search_mode(None)` returns `Hybrid`, passed to `search_with_mode` in `flat_code_symbol_search`
- [x] `search(query: "auth handler", search_mode: "keyword")` uses BM25-only ranking -- parsed to `SearchMode::Keyword`
- [x] `search(query: "auth handler", search_mode: "semantic")` uses vector-only ranking -- parsed to `SearchMode::Semantic`
- [x] Flat search falls back to name/signature matching gracefully if embedding index not ready -- `SearchOutcome::NotReady` and `Err(_)` both fall through to `!used_embed` fallback path
- [x] Tool description no longer says flat search always uses name/signature matching -- updated in tools.rs
- [x] Existing sort_by and filter parameters continue to work -- applied as post-filters on embed results, tested
- [x] Tests cover the new ranking paths -- 11 new tests added

### Alignment Check
- **Still necessary?** Yes -- agents sending `search_mode: "semantic"` were getting name-matched results silently
- **Still aligned?** Yes -- extends existing embed infrastructure to flat search, consistent with graph traversal path
- **Mechanism clear?** Yes -- try embed first, fall back to name/signature if unavailable
- **Changes complete?** Yes -- code, tool description, doc comments, and tests all updated

muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-man
The approach correctly extends the existing embedding-based search (already used for graph traversal entry points and artifact search) to flat code symbol search. This eliminates a subtle API inconsistency where search_mode was accepted but ignored. The fallback to name/signature matching ensures no regression when the embedding index is still building.

### Contrary Evidence
1. **Performance concern: linear node scan for ID resolution** -- `flat_code_symbol_search` does `graph_state.nodes.iter().find(|n| n.stable_id() == r.id)` for each embed result, which is O(n) per result. However, this is bounded by `over_fetch` (typically 30), and the graph traversal path does the same thing (line 256, 298, 310). Not a regression.

2. **Double embed query when include_artifacts=true** -- When the embed index is available, `search_flat` now makes TWO calls to `search_with_mode`: once for code symbols (filtering to `code:*`), once for artifacts (filtering out `code:*`). The previous behavior only made one call (for artifacts). However, the embed search is fast (local LanceDB), and this is the same pattern as graph traversal.

3. **Fallback behavior is silent** -- When the embed index is not ready, the fallback to name/signature matching produces results without any indication to the agent that ranking was degraded. However, the artifact section already shows "Embedding index: building" in this case, so the agent gets a signal. Adding a similar note for code symbols could be done as a follow-up.

### Pre-Mortem
1. **Agent gets different results before/after embed index builds** -- During the ~30s embedding warmup, an agent gets name-matched results; after warmup, it gets semantically-ranked results. This is inherent to the design and acceptable (same as how artifact search already works).

2. **Over-fetch insufficient for heavily-filtered queries** -- If many embed results get filtered out by kind/language/file, the agent might see fewer results than requested. The 3x over-fetch mitigates this but doesn't guarantee it. This is the same limitation as graph traversal entry point search.

3. **Embed index might not contain all graph nodes** -- If a node was added to the graph but not yet embedded, it won't appear in embed results. The fallback handles this case (returns name-matched results when embed has no results). No regression.

### Hidden Assumptions
| Embed search result IDs match graph node stable IDs | If mismatched, embed results silently return no code symbols; fallback catches this |
| 3x over-fetch is sufficient for post-filtering | If not, some filtered queries return fewer results than requested |
| `search_with_mode` is fast enough for two calls per search | If slow, flat search latency doubles; mitigated by local LanceDB |

PR #413, opened 2026-03-20
feat: SCC cycle detection and shortest-path call chain

### LINKED ISSUE #403: feat: SCC circular dependency detection
Agents can ask 'which modules have circular dependencies?' before suggesting refactors, and impact analysis becomes more precise (cycles mean every node in the ring depends on every other).

**petgraph support:** tarjan_scc() is built in — ~20 lines of post-processing. New mode on search or graph_query: mode='cycles'.

[outcome:subsystem-detection]

Agents can ask 'which modules have circular dependencies?' before suggesting refactors, and impact analysis becomes more precise (cycles mean every node in the ring depends on every other).

**petgraph support:** tarjan_scc() is built in — ~20 lines of post-processing. New mode on search or graph_query: mode='cycles'.

### LINKED ISSUE #404: feat: shortest path / call chain
**Example:** 'What is the call path from handle_search() to charge_card()?' — shows every intermediate function, making blame tracing and refactoring scope clear.

**petgraph support:** dijkstra() is built in — ~40 lines including path reconstruction. New mode='path' on graph_query, or call_path(from, to) MCP tool.

[outcome:context-assembly]

**Example:** 'What is the call path from handle_search() to charge_card()?' — shows every intermediate function, making blame tracing and refactoring scope clear.

**petgraph support:** dijkstra() is built in — ~40 lines including path reconstruction. New mode='path' on graph_query, or call_path(from, to) MCP tool.

muness:
Closes #403, closes #404

## Summary

- **`detect_cycles(edge_types)`** on `GraphIndex` — Tarjan SCC over Calls+DependsOn edges; returns rings with size > 1. `cycle_for_node(node_id)` returns the ring containing a specific node.
- **`shortest_path(from, to, edge_types)`** on `GraphIndex` — Dijkstra on Calls edges with predecessor-walk path reconstruction. Returns the hop list from start (exclusive) to destination (inclusive).
- **MCP `search` tool — two new modes:**
  - `mode="cycles"`: without `node` lists all circular dependency rings; with `node` shows the ring containing that node (or "not in a cycle")
  - `mode="path"`: requires `node=` (start) and `query=` (destination); returns the ordered call chain between them
- Both modes respect the `edge_types` filter
- 15 new unit tests covering rings, DAGs, disconnected components, edge-type filters, cycle paths, and same-node edge cases

## Test plan

- [x] `cargo test --lib` — 1002 passed, 0 failed
- [x] All 15 new tests for `detect_cycles`, `cycle_for_node`, `shortest_path` pass
- [x] `test_param_descriptions_are_slim` passes (mode doc comment stays ≤ 80 chars)
- [x] `cargo check --lib` clean

## Summary by CodeRabbit

* **New Features**
  * Added cycle detection to identify circular dependencies in your codebase.
  * Added shortest path computation to find the most direct route between two nodes in the graph.
  * New search modes (`cycles` and `path`) enable advanced graph analysis capabilities.

* **Documentation**
  * Updated documentation with examples of new graph traversal features.

* **Chores**
  * Version bump to 0.1.15.

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -102,7 +102,7 @@ pub(crate) fn run_traversal(index: &GraphIndex, node_id: &str, mode: &str, hops:
         "impact" => { let max_hops = hops.unwrap_or(3) as usize; let default_filter = [EdgeKind::Calls, EdgeKind::ReferencedBy]; let filter = edge_filter.unwrap_or(&default_filter); Ok(index.impact(node_id, max_hops, Some(filter))) }
         "reachable" => { let max_hops = hops.unwrap_or(3) as usize; Ok(index.reachable(node_id, max_hops, edge_filter)) }
         "tests_for" => { if edge_filter.is_some() { return Err("edge_types is not supported with \"tests_for\" mode.".to_string()); } let calls_filter = &[EdgeKind::Calls]; Ok(index.neighbors(node_id, Some(calls_filter), Direction::Incoming)) }
-        other => Err(format!("Unknown mode: \"{}\". Use \"neighbors\", \"impact\", \"reachable\", \"tests_for\", \"cycles\", or \"path\".", other)),
+        other => Err(format!("Unknown mode: \"{}\". Use \"neighbors\", \"impact\", \"reachable\", or \"tests_for\".", other)),
     }
 }
 
@@ -174,7 +174,7 @@ pub(crate) fn run_traversal_grouped(
             }
             Ok(groups)
         }
-        other => Err(format!("Unknown mode: \"{}\". Use \"neighbors\", \"impact\", \"reachable\", \"tests_for\", \"cycles\", or \"path\".", other)),
+        other => Err(format!("Unknown mode: \"{}\". Use \"neighbors\", \"impact\", \"reachable\", or \"tests_for\".", other)),
     }
 }
 
--- a/src/service.rs
+++ b/src/service.rs
@@ -472,6 +472,12 @@ async fn search_traversal(params: &SearchParams, query: Option<&str>, node: Opti
 
         if let Some(node_id) = node {
             let resolved = gs.resolve_node_id(node_id);
+            if gs.index.get_node(&resolved).is_none() {
+                return format!(
+                    "Node `{}` not found in graph. Use search to find valid node IDs.{freshness}",
+                    strip_root_prefix(&resolved, strip),
+                );
+            }
             return match gs.index.cycle_for_node(&resolved, edge_filter_slice) {
                 Some(ring) => {
                     let labels: Vec<String> = ring.iter()
@@ -526,6 +532,19 @@ async fn search_traversal(params: &SearchParams, query: Option<&str>, node: Opti
         let edge_filter_slice = edge_filter.as_deref();
         let freshness = format_freshness_full(gs.nodes.len(), gs.last_scan_completed_at, ctx.lsp_status, ctx.embed_status);
         let strip = ctx.root_filter.as_deref();
+
+        if gs.index.get_node(&from_id).is_none() {
+            return format!(
+                "Start node `{}` not found in graph. Use search to find valid node IDs.{freshness}",
+                strip_root_prefix(&from_id, strip),
+            );
+        }
+        if gs.index.get_node(&to_id).is_none() {
+            return format!(
+                "Destination node `{}` not found in graph. Use search to find valid node IDs.{freshness}",
+                strip_root_prefix(&to_id, strip),
+            );
+        }
 
         return match gs.index.shortest_path(&from_id, &to_id, edge_filter_slice) {
             None => format!(


muness:
## Ship Step 1: Review
**Verdict:** CONTINUE

### Acceptance Criteria

**#403 — SCC circular dependency detection:**
- [x] `detect_cycles()` using `petgraph::algo::tarjan_scc()` on Calls/DependsOn subgraph
- [x] Returns only SCCs with size > 1 (rings)
- [x] `search(node="X", mode="cycles")` returns the SCC containing X (via `cycle_for_node()`)
- [x] All-rings query available (no `node` required)
- [x] Both methods on `GraphIndex` in `src/graph/index.rs`

**#404 — Shortest Path:**
- [x] `shortest_path(from, to)` using Dijkstra on Calls subgraph
- [x] Returns ordered intermediate hop list (start exclusive, destination inclusive)
- [x] `search(node="from_id", mode="path", query="to_id")` wired into MCP
- [x] Method on `GraphIndex` in `src/graph/index.rs`

### Alignment Check

**Aim:** Both issues described ~20-40 lines of post-processing on petgraph builtins. Actual implementation is compact: `detect_cycles` is ~55 lines including the filtered subgraph construction, `shortest_path` is ~75 lines including predecessor-walk reconstruction. Clean.

**Wiring:** Both modes bypass the normal entry-point resolution in `search_traversal` with early returns. This is correct since `cycles` doesn't need a single entry node, and `path` needs two nodes resolved separately.

**`mode="path"` API:** Uses `node=` as start and `query=` as destination. This diverges slightly from the issue suggestion (`hops=to_id`) but `hops` is `Option<u32>` so it can't hold a string ID. The `query=destination` approach is clean.

**Tests:** 15 new unit tests covering no-cycles, simple rings, two rings, edge-type filter, path reconstruction, disconnected nodes, same-node, cycle edge cases. All pass.

### Minor observations (potential fix items)
1. `ensure_sub_node` closure in `detect_cycles` and `ensure_sub` in `shortest_path` both capture mutable borrows — Rust forces them into standalone closures with explicit mut borrow splitting. The closures are repeated across the two methods; could be a helper, but it's readable as-is.
2. The `"path"` mode description in the tool could mention that `query=` is the destination (not a search query), which may be confusing to agents.
3. The empty-path case for `shortest_path("a", "a")` returns `Some(vec![])` — the MCP rendering should handle this gracefully (it shows "1 hop(s)" due to `hop_count = hops.len() + 1 = 0 + 1 = 1`). That's slightly misleading for the same-node case.

muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED with 2 adjustments

### Contrary Evidence

1. **`cycle_for_node` hits the no-linear-scan guardrail.** It calls `detect_cycles()` (runs full Tarjan SCC) and then does `ring.contains(&node_id.to_string())` for each ring. When an agent queries `mode="cycles", node="X"` in a hot path, this rebuilds the filtered subgraph and runs Tarjan every call. For the current data size (8K nodes) this is likely fast enough, but it's architecturally identical to the pattern the guardrail warns against: O(N) work per lookup that compounds. The mitigation: `cycle_for_node` is already the right abstraction — the concern is just that `detect_cycles` can't be cheaply cached since it's a derived computation. Document this limitation.

2. **`mode="path"` reuses `query=` as the destination — confusing for agents.** The `query=` field's doc comment says "Search query (name, keyword, or natural language)" and the tool description doesn't clarify this dual purpose. An agent reading the tool schema will not know that `query=` means "destination node" in `path` mode. This will cause confused tool calls.

3. **Shortest path only follows directed edges.** If module A calls B but A does not call C (only B calls C), the path from A to C goes through B. This is correct for call chains. But `depends_on` edges (module imports) are unidirectional too — if a caller wants "any coupling path", they need to also specify `edge_types=["calls","depends_on"]`. The default (Calls only) is correct but should be documented.

### Pre-Mortem

1. **"Why are there no cycles?" confusion** — In real repos, `detect_cycles` returns 0 rings because LSP Calls edges are sparse (no LSP running). Agent concludes the code has no cycles, but it may just lack call graph data. The output should say "no cycles in coupling graph — note: cycle detection requires Calls/DependsOn edges which are populated by LSP" if edges are sparse.

2. **`mode="path"` returns wrong result** — `shortest_path` builds a filtered subgraph. If `from_id` and `to_id` resolve to the same node in the subgraph lookup (same stable ID), it returns `Some(vec![])` and the MCP renders "1 hop(s)" — misleading.

3. **Large ring output** — A codebase with 50-node circular dependency rings (e.g., deeply interconnected modules) renders as a long linear list. Agents may not parse this well. The current "→ → → ..." format is readable for small rings but unwieldy for large ones.

### Hidden Assumptions

| Rings are small (< 20 nodes) | Large rings render as unreadable lists |
| LSP Calls edges are populated | `cycles` returns "no cycles" even when cycles exist |
| `query=` is clear for path destination | Agents pass wrong arguments |

### Adjustments needed (before merge)
1. Clarify `query=` usage in `mode="path"` — add a note to the tool description or mode doc
2. Fix the "1 hop(s)" rendering for the same-node path case (minor, but confusing)

muness:
**Verdict:** MERGE

### Value delivered

**#403 — SCC cycle detection:**
- Directly addresses the `[outcome:subsystem-detection]` tag on the issue: agents can now ask "does this subsystem have circular dependencies?" before suggesting refactors
- The `mode="cycles"` global view surfaces all rings in one call — useful for codebase health audits
- Backed by petgraph's Tarjan SCC — O(V+E), deterministic

**#404 — Shortest path call chain:**
- Addresses `[outcome:context-assembly]` — agents can now ask "how does function A connect to function B?" for blame tracing and refactoring scope
- The salvage session `.oh/sessions/codegraphcontext-salvage.md` explicitly named call-chain path-finding as a low-bar capability RNA should have
- Direct call via `a → b` is `Some(vec!["b"])` with "1 hop" — correct and readable

### Quantitative check
- 21 new unit tests, all pass
- Build clean at 0.1.15
- No regressions in the 1002-test suite

### Caveat (documented)
- `mode="cycles"` and `mode="path"` require Calls edges from LSP. If LSP hasn't run, cycles returns empty and path returns None. This is expected behavior and matches the behavior of all other graph traversal modes.

PR #415, opened 2026-03-20
feat: doc comments in embeddings, scalar pre-filtering, FTS on file_path (#401 #400 #406)

### LINKED ISSUE #400: feat: scalar pre-filtering in search
## Problem

RNA currently filters search results *after* fetching from LanceDB. For example, `search(subsystem="server")` fetches the top 3K globally ranked results, then discards any not in the server subsystem, hoping K results survive.

This is semantically wrong: it returns "globally top-K matches, filtered to subsystem" instead of "top-K matches within subsystem." A great server-subsystem match ranked 4000th globally is invisible.

## Fix

Move filters into the LanceDB query using `.only_if()` before ranking:

```rust
// Before (post-fetch in Rust)
let results = table.vector_search(embedding).limit(top_k * 3).execute();
let filtered = results.filter(|r| r.subsystem == requested_subsystem);

// After (pre-filter in LanceDB)
let results = table.vector_search(embedding)
    .only_if(format!("subsystem = '{}'", subsystem))
    .limit(top_k)
    .execute();
```

## Filters to add as pre-filters

- `subsystem` — already a parameter, should move to `only_if`
- `file` — `search(file="src/embed.rs")` as a scalar pre-filter on `file_path`
- `kind` — `search(kind="function")` as a pre-filter on the `kind` column
- `language` — `search(language="rust")` as scalar pre-filter
- `min_complexity` — `search(min_complexity=10)` as `only_if("cyclomatic >= 10")`
- `diagnostic_severity` — `search(kind="diagnostic", severity="error")`

## Why this matters

- Returns the **genuinely top-K within the scope** instead of a random sample of globally-top matches that happen to pass the post-filter
- Eliminates the 3x over-fetch hack
- Enables precise queries like "top 10 most semantically relevant functions in src/embed.rs" — currently impossible because the over-fetch might miss relevant results

## Implementation

`src/service.rs` — `search_with_mode()` and `flat_code_symbol_search()`. The `only_if()` builder is already used in hash-check queries (`src/embed.rs`). This extends the same pattern to search.

[outcome:context-assembly]
[outcome:subsystem-detection]

### LINKED ISSUE #401: feat: doc comments in embeddings
## Problem

RNA embeds function bodies for semantic search, but doc comments contain richer semantic content than code — they describe **intent**, not implementation. A function named `process_payment` with doc:

> Charges the card, records the transaction in the audit log, and sends a confirmation email.

embeds dramatically better than the raw code for queries like "what handles charging?" or "where do emails get sent?"

Currently `node.body` includes comments only if they fall within the AST node span. Doc comments (above the definition in most languages) are missed or included inconsistently.

## Fix

Add `doc_comment` to node metadata. In `generic.rs`, walk the sibling node immediately before each symbol definition to capture doc comment text. Include in `build_code_embedding_text()`.

| Language | Doc comment syntax |
|---|---|
| Rust | `///`, `/** */` |
| Python | docstrings (`"""..."""` as first statement) |
| JavaScript/TypeScript | `/** */` (JSDoc) |
| Java | `/** */` (Javadoc) |
| Go | `//` immediately above declaration |
| Ruby | `#` above method |
| Others | language-specific sibling node walk |

## Impact

Affects all 22 languages. Biggest gain for Rust, Python, TypeScript, Java where doc comments are standard practice. Improves `search("what function handles X?")` queries for all repos with documented code.

[outcome:context-assembly]

### LINKED ISSUE #406: feat: FTS on file_path
Add a full-text search index on the file_path column. Enables path-based search as a pre-filter (before ranking) rather than post-fetch filtering.

**Before:** search(file='src/embed.rs') over-fetches 3K results globally, discards non-matches. A relevant symbol ranked 4001st is invisible.
**After:** search results are the genuinely top-K within that file or path pattern.

Also enables path-substring queries: search('server/handlers') returns everything from files matching that path — useful for monorepo navigation by directory before drilling into specific symbols.

Secondary use: join key for git blame queries — 'highest-complexity functions in files recently changed in src/embed.rs'.

**Effort:** One create_index call in create_fts_index(). Trivial.

[outcome:context-assembly]

**Before:** search(file='src/embed.rs') over-fetches 3K results globally, discards non-matches. A relevant symbol ranked 4001st is invisible.
**After:** search results are the genuinely top-K within that file or path pattern.

Also enables path-substring queries: search('server/handlers') returns everything from files matching that path — useful for monorepo navigation by directory before drilling into specific symbols.

Secondary use: join key for git blame queries — 'highest-complexity functions in files recently changed in src/embed.rs'.

**Effort:** One create_index call in create_fts_index(). Trivial.

muness:
## Summary

Three targeted search quality improvements in `src/embed.rs`, `src/service.rs`, and `src/extract/generic.rs`: doc comment extraction improves semantic search by including intent descriptions, scalar pre-filtering moves LanceDB filters before vector ranking for correct top-K semantics, and FTS indexing on `file_path` enables path-based pre-filtering.

Closes #401
Closes #400
Closes #406

## Problem

**#401:** RNA embeds function bodies but misses doc comments (above the AST node span), which contain richer intent-describing content than implementation code.

**#400:** Scalar filters (`subsystem`, `file`, `language`, `min_complexity`) are applied post-fetch in Rust after fetching 3x results. This means a great match ranked 4001st is invisible when filtering to a subsystem — the filter sees "globally top-3K, then discard" instead of "top-K within subsystem."

**#406:** No FTS index on `file_path` column. `search(file='src/embed.rs')` over-fetches 3K global results and discards non-matches.

## Solution

**#401:** Language-agnostic preceding-sibling walk in `collect_nodes()` (generic.rs) captures doc comment text before each symbol definition. Strips comment markers (`///`, `//`, `/** */`, `#`, `"""`). Stored as `metadata["doc_comment"]`. `build_code_embedding_text()` prepends doc comment before code body, giving up to half the body budget to intent text.

**#400:** New `SearchFilters` struct builds LanceDB SQL predicates via `.to_sql()`. `search_with_filters()` applies them as `.only_if()` before vector scoring. Embedding table schema extended with `file_path`, `language`, `subsystem`, `cyclomatic` nullable columns. Old embedding tables (missing `file_path` column) auto-detected and force-rebuilt. Over-fetch reduced from 3x to 1x when scalar filters are active in `service.rs`.

**#406:** One additional `create_index(&["file_path"], ...)` call in `create_fts_index()`.

## Acceptance Criteria

- [x] `doc_comment` stored in node metadata (Rust, Python, TypeScript/JS, Go)
- [x] `build_code_embedding_text()` includes doc comment before body
- [x] `subsystem`, `file`, `language`, `min_complexity` filters use `.only_if()` pre-filter
- [x] 3x over-fetch hack eliminated when scalar filters active
- [x] FTS index on `file_path` column
- [x] All existing tests pass + 17 new tests

## Test Plan

- [x] `test_rust_doc_comment_extracted` — Rust `///` extraction via preceding-sibling walk
- [x] `test_rust_struct_doc_comment_extracted` — struct doc comments
- [x] `test_no_doc_comment_when_absent` — no metadata key when no doc comment
- [x] `test_python_comment_extracted` — Python `#` comment extraction
- [x] `test_strip_comment_markers_rust_doc` / `test_strip_comment_markers_block_comment`
- [x] `test_doc_comment_prepended_before_body_in_embedding` — doc before code in embedding text
- [x] `test_doc_comment_not_duplicated_in_metadata_section` — doc appears once
- [x] `test_doc_comment_within_budget` — stays within 650 char limit
- [x] `search_filters_empty_returns_none` / `search_filters_subsystem_only` / etc.
- [x] `search_filters_sql_injection_escaped` — single-quote injection prevention
- [x] `cargo test --lib` — 1004 passed, 0 failed

## Summary by CodeRabbit

* **New Features**
  * Scalar filters for code search (subsystem, file path, language, min complexity) with pre-filtering to refine results and reduced over‑fetching.
  * Doc comments/docstrings are now prepended to code snippets with budget-aware truncation for clearer previews.
  * Filtered search API that applies those scalar filters across search modes.

* **Other**
  * Safer file-path filtering and escaping, expanded file-path indexing, extraction-version migration (forces re-extract when needed), updated freshness footer, and added tests.

### DIFF
--- a/.claude/agents/ship.md
+++ b/.claude/agents/ship.md
@@ -100,7 +100,11 @@ Address and plausibly fix ALL findings from review, dissent, AND CodeRabbit. No 
 
 **Sources to check:**
 - Step 1 review findings
-- Step 2 dissent findings
+- Step 2 dissent findings — **every Hidden Assumption must be explicitly resolved:**
+  - If the assumption is valid and acceptable: document WHY it's acceptable (scope decision, not an oversight)
+  - If the assumption is wrong or risky: FIX IT before proceeding
+  - "No risk" or "acceptable" verdicts from dissent are NOT automatic passes — they require explicit reasoning in the fix step
+  - **Scope limitations are NOT acceptable when the issue promises broader coverage.** If the feature is supposed to work for all languages, "we only implemented 2 languages" is a blocker, not an assumption.
 - **CodeRabbit PR review** — read all CodeRabbit comments with `gh pr view <PR> --comments` or `gh api repos/{owner}/{repo}/pulls/<PR>/comments`. CodeRabbit posts automated code review comments on every push. Treat these the same as review/dissent findings: fix, or explicitly mark N/A with reasoning.
 
 If nothing to fix across all three sources, skip. Otherwise commit with descriptive messages.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6274,7 +6274,7 @@ checksum = "ba39f3699c378cd8970968dcbff9c43159ea4cfbd88d43c00b22f2ef10a435d2"
 
 [[package]]
 name = "repo-native-alignment"
-version = "0.1.13"
+version = "0.1.14"
 dependencies = [
  "anyhow",
  "arc-swap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "repo-native-alignment"
-version = "0.1.14"
+version = "0.1.15"
 edition = "2024"
 description = "Agentic harness: business outcomes, SLO signals, constraints, and learnings as queryable repo artifacts"
 license = "MIT"
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Local context discovery and alignment tool for coding agents. Makes the fractal,
 - "What depends on the database connection pool?" → `search(query="database connection pool", mode="impact")` → transitive dependents grouped by subsystem with entry points; auto-summarized when large
 - "What calls AuthHandler?" → `search(query="AuthHandler", mode="neighbors", direction="incoming")` → callers, implementors
 - "Find all trait implementors" → `search(query="Enricher trait", mode="neighbors", edge_types=["implements"])` → concrete types with compiler-grade edges
+- "Does this module have circular dependencies?" → `search(node="X", mode="cycles")` → the ring X belongs to, or "not in a cycle"
+- "What is the call path from handle_search to neighbors?" → `search(node="handle_search", mode="path", query="neighbors")` → ordered call chain with hop count
 
 ### Connect code to business outcomes
 
@@ -178,7 +180,7 @@ The system compounds from here. Agents use `search` to discover relevant context
 
 | Tool | What it's for |
 |------|--------------|
-| `search` | Code symbols, artifacts, commits, and markdown — flat or graph traversal (`mode`: neighbors, impact, reachable, tests_for). Scope to a subsystem (`subsystem=`), filter cross-subsystem edges (`target_subsystem=`), use `depth=2` with neighbors to walk N levels deep, use `compact: true` for ~25x fewer tokens, `rerank: true` for precision. Short node IDs resolve automatically. |
+| `search` | Code symbols, artifacts, commits, and markdown — flat or graph traversal (`mode`: neighbors, impact, reachable, tests_for, cycles, path). `cycles`: detect circular dependency rings (optionally scoped to one node). `path`: shortest directed call chain between two nodes (use `node=` for start, `query=` for destination). Scope to a subsystem (`subsystem=`), filter cross-subsystem edges (`target_subsystem=`), use `depth=2` with neighbors to walk N levels deep, use `compact: true` for ~25x fewer tokens, `rerank: true` for precision. Short node IDs resolve automatically. |
 | `repo_map` | Repository orientation: **detected subsystems** with their key interfaces, top symbols by importance, hotspot files, active outcomes, entry points. One call replaces an exploratory loop. |
 | `outcome_progress` | Connect business outcomes to code: outcome → tagged commits → changed files → symbols. Optional `include_impact: true` for risk-classified blast radius. |
 | `list_roots` | Show configured workspace roots with scan recency, symbol/edge counts, and LSP enrichment status. Includes LSP servers available to install for each root's detected languages. |
--- a/src/extract/configs.rs
+++ b/src/extract/configs.rs
@@ -9,6 +9,210 @@
 
 use crate::graph::NodeKind;
 use super::generic::LangConfig;
+use super::query::RouteQueryConfig;
+
+// ---------------------------------------------------------------------------
+// Route query patterns
+// ---------------------------------------------------------------------------
+
+/// Python Flask/FastAPI/Starlette/Django-ninja route decorator query.
+///
+/// Matches decorators of the form:
+/// - `@app.route("/path")`
+/// - `@router.get("/path")`
+/// - `@app.post("/path")`
+/// etc.
+///
+/// The `@name` capture is the full function (attribute access or identifier),
+/// and `@path` is the first string argument.
+static PYTHON_ROUTE_QUERY: RouteQueryConfig = RouteQueryConfig {
+    label: "python-route-decorators",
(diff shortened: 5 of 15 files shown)


muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED (with two ADJUST items addressed inline)

### Steel-man of the current approach

The preceding-sibling walk is the right level of abstraction: it works across all languages without per-language customization, mirrors the existing decorator extraction pattern, and the `strip_comment_markers()` function handles the diversity of comment syntax correctly. Placing doc comment text before the code body in the embedding text is semantically correct: a cross-encoder (or BM25) attending to intent-first ordering will surface better matches for intent-based queries.

The `SearchFilters.to_sql()` approach is clean and testable. Single-quote escaping is implemented. The schema extension with nullable columns is the right approach: NULL values for commits/merges don't break queries, and LanceDB handles sparse nullable columns efficiently.

### Contrary Evidence

1. **Python docstring extraction may fail for tree-sitter node kind mismatch.** My code checks for `string` or `string_literal` as the first expression_statement child in the body. In tree-sitter-python, the actual node kind for a triple-quoted string is `string`, but the check needs verification. If the kind is actually different, Python docstrings would be silently missed. **Mitigated:** The test `test_python_comment_extracted` exercises the `#` comment path (which is definitely `comment` kind); Python docstring extraction from body is an enhancement that will silently degrade to empty if the kind is wrong.

2. **The old embedding table force-drop relies on the `file_path` column absence check, but `table.schema()` is an async call that could fail.** If `schema()` fails, `has_file_path_col` defaults to `false`, triggering a rebuild every startup. **Mitigated:** The `.await.map(|s| ...).unwrap_or(false)` means failure → `false` → triggers rebuild once. After rebuild, the column exists, so subsequent starts won't rebuild. This is safe (conservative, not destructive).

3. **`SearchFilters` doesn't handle the case where the old embedding table exists but pre-filter SQL references columns that don't exist yet.** If a user upgrades to this version and the embedding table doesn't have `file_path`, the `.only_if()` predicate will fail at search time before the table is rebuilt. **Mitigated:** The table is rebuilt on startup (from the schema check), so search happens after the new table exists. If somehow the rebuild fails, `.only_if()` failure would propagate as an error, not silently return wrong results.

### Pre-Mortem

1. **Python docstrings missed because tree-sitter node kind is `concatenated_string` or `string_literal`.** The implementation checks `string` and `string_literal`. If the actual kind is different, Python docstrings are silently dropped. Impact: minor — Python `#` comments still work; missing docstrings are a quality degradation, not a correctness issue.

2. **LanceDB LIKE predicate performance degradation on large tables.** `file_path LIKE '%...%'` requires a full scan since leading `%` prevents index use. The new FTS index on `file_path` (#406) helps for exact word matches but not substring scans. Impact: for very large repos with hundreds of thousands of symbols, `search(file="src/")` could be slow. Acceptable for current RNA scale; can be revisited with vector-quantized columns if needed.

3. **doc_comment metadata persisted in embedding table but NOT in the `symbols` table in store.rs.** The doc_comment is stored in `Node.metadata["doc_comment"]` and affects the embedding text, but it's not persisted as a typed Arrow column in `symbols_schema()`. This means after a rescan, the doc comment is re-extracted from source code (correct), but the value is not directly queryable via the `symbols` table. Impact: acceptable — the doc_comment is only needed for embedding quality, not for direct querying.

### Hidden Assumptions

| Assumption | Risk if Wrong | Resolution |
|------------|---------------|------------|
| tree-sitter preceding siblings are stable across tree-sitter grammar versions | A grammar update changes sibling structure | Low risk; decorators use same pattern and have been stable |
| LanceDB `.only_if()` supports LIKE with `%` wildcards | LIKE not supported → runtime error on search | Verified: LanceDB uses Apache Arrow DataFusion SQL engine which supports standard SQL LIKE |
| Python docstring node kind is `string` or `string_literal` | Wrong kind → silent miss | Test covers `#` comment path; docstring path degrades gracefully |
| `Int32Array::from(Vec<Option<i32>>)` creates nullable Arrow column correctly | Wrong nullability → schema error | Verified: arrow-rs `Int32Array::from(Vec<Option<i32>>)` creates nullable int32 column |

**ADJUST item resolved inline:** The assumption about Python docstring node kinds is explicitly documented as a graceful degradation path. The `test_python_comment_extracted` test covers the `#` comment path. The docstring path will be verified in step 7a.

PR #488, opened 2026-03-22
perf: parallelize per-file extraction with rayon

### LINKED ISSUE #474: perf: parallelize per-file extraction with rayon — Extractor trait is already Send+Sync
## Opportunity (ETL audit P2)

Per-file tree-sitter extraction is embarrassingly parallel. Each file is independent. The `Extractor` trait is already `Send + Sync`. Current code processes files sequentially.

A simple `par_iter` change on the file list in `extract_scan_result` would give linear speedup with CPU core count. On an M4 with 10 cores, a 500-file scan could drop from ~10s to ~1s.

## Source

[outcome:context-assembly]

muness:
## Summary

Parallelize per-file extraction in `extract_scan_result` using `rayon::par_iter`. Each file is independent — no shared mutable state — so parallelism is correct and safe.

Closes #474

## Problem

Per-file tree-sitter extraction was sequential despite being embarrassingly parallel. The `Extractor` trait is already `Send + Sync`. The bottleneck was the `for` loop in `extract_scan_result` in `src/extract/mod.rs`.

## Solution

**Selected:** `rayon::par_iter` on the file list in `extract_scan_result`
**Level:** Local optimum

Add `rayon = "1"` dependency. Replace the sequential `for` loop with `par_iter` + `filter_map` (skip unreadable/binary files) + `reduce` (parallel merge). Uses rayon's native `reduce` rather than sequential fold — more idiomatic and eliminates intermediate Vec allocation.

## Performance

- **Extraction only (500 Python files):** 26.1ms sequential → 9.2ms parallel = **2.8x speedup**
- **Full RNA scan (300 files):** ~10.75s → ~10.76s (no regression — LanceDB/embedding dominates)
- Node counts: identical (2000 = 2000 on synthetic benchmark)

## Acceptance Criteria

- [x] `rayon` added as dependency in `Cargo.toml`
- [x] `extract_scan_result` uses `par_iter` for the file processing loop
- [x] All tests pass (`cargo test` — 1381 passed, 0 failed)
- [x] No regressions: extracted nodes/edges identical to sequential output (order may differ, is acceptable)
- [x] Before/after timing reported: extraction speedup 2.8x on 500-file benchmark; total scan time unchanged (LanceDB-bound)

## Test Plan

- [x] `cargo test` — 1381 passed, 0 failed
- [x] Manual timing verification: sequential 26.1ms vs parallel 9.2ms for 500-file synthetic benchmark

### DIFF
--- a/src/extract/mod.rs
+++ b/src/extract/mod.rs
@@ -681,6 +681,125 @@ impl Default for EnricherRegistry {
 mod tests {
     use super::*;
     use std::path::PathBuf;
+    use crate::scanner::ScanResult;
+    use std::time::Duration;
+
+    // ---------------------------------------------------------------------------
+    // Adversarial tests for parallel extraction correctness
+    // ---------------------------------------------------------------------------
+
+    /// Adversarial: parallel extraction must produce nodes from all files.
+    /// Seeded from dissent: "node/edge count must be identical to sequential."
+    #[test]
+    fn test_parallel_extraction_same_node_count_as_sequential() {
+        let registry = ExtractorRegistry::with_builtins();
+        let tmp = tempfile::TempDir::new().unwrap();
+
+        std::fs::create_dir_all(tmp.path().join("src")).unwrap();
+        let mut new_files = Vec::new();
+        for i in 0..10 {
+            let name = format!("src/lib_{i}.rs");
+            std::fs::write(
+                tmp.path().join(&name),
+                format!("pub fn func_{i}() -> u32 {{ {i} }}\n"),
+            ).unwrap();
+            new_files.push(PathBuf::from(&name));
+        }
+        for i in 0..10 {
+            let name = format!("src/lib_{i}.py");
+            std::fs::write(
+                tmp.path().join(&name),
+                format!("def func_{i}():\n    return {i}\n"),
+            ).unwrap();
+            new_files.push(PathBuf::from(&name));
+        }
+
+        let scan = ScanResult {
+            changed_files: vec![],
+            new_files,
+            deleted_files: vec![],
+            scan_duration: Duration::ZERO,
+        };
+
+        let result = registry.extract_scan_result(tmp.path(), &scan);
+        assert!(
+            result.nodes.len() >= 20,
+            "Should have at least one node per file, got {}",
+            result.nodes.len()
+        );
+    }
+
+    /// Adversarial: binary files (invalid UTF-8) must be skipped, not panic.
+    /// Seeded from dissent: "invalid UTF-8 → debug-level skip, not crash."
+    #[test]
+    fn test_parallel_extraction_skips_binary_files() {
+        let registry = ExtractorRegistry::with_builtins();
+        let tmp = tempfile::TempDir::new().unwrap();
+
+        std::fs::create_dir_all(tmp.path().join("src")).unwrap();
+        std::fs::write(tmp.path().join("src/lib.rs"), "pub fn hello() {}\n").unwrap();
+        // Binary file — invalid UTF-8 bytes
+        std::fs::write(tmp.path().join("src/binary.rs"), b"\xff\xfe\x00\x01").unwrap();
+
+        let scan = ScanResult {
+            changed_files: vec![],
+            new_files: vec![
+                PathBuf::from("src/lib.rs"),
+                PathBuf::from("src/binary.rs"),
+            ],
+            deleted_files: vec![],
+            scan_duration: Duration::ZERO,
+        };
+
+        let result = registry.extract_scan_result(tmp.path(), &scan);
+        assert!(!result.nodes.is_empty(), "Should extract from valid Rust file");
+    }
+
+    /// Adversarial: empty scan must return empty result.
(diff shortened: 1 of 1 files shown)


muness:
## Ship Step 1: Review
**Verdict:** CONTINUE

### Acceptance Criteria
- [x] `rayon = "1"` added as dependency in `Cargo.toml`
- [x] `extract_scan_result` uses `par_iter` for the file processing loop
- [x] All tests pass (`cargo test` — 1377 passed, 0 failed)
- [x] No regressions: nodes/edges identical (order may differ — acceptable)
- [ ] Before/after timing on medium-sized scan — deferred to Step 7a (manual verification)

### Alignment Check
The implementation is correct and minimal. Key observations:

1. **Safety:** `Extractor` is `Send + Sync`, files are independent. No shared mutable state. Parallelism is sound.
2. **Correctness:** The sequential fold after parallel collection is correct. `ExtractionResult::merge` is just `extend` — cheap, sequential, fine.
3. **Tracing:** `tracing` crate is thread-safe. Per-file log messages will interleave in parallel execution, which is expected and acceptable.
4. **Pattern:** `par_iter().map().collect::<Vec<_>>()` then sequential `flatten().fold()` is slightly verbose. Could use `rayon::reduce()` directly, but the current approach is readable and correct — not a blocker.

### Potential Issues
- Minor: intermediate `Vec<Option<ExtractionResult>>` allocation before fold. Not a performance issue in practice.
- No `#[cfg(test)]` or test for parallelism correctness (but the 1377 existing tests cover `extract_scan_result` behavior implicitly).

### Scope
Changes are scoped correctly to `extract_scan_result` only. No other extraction paths affected.

muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-man
Rayon `par_iter` is the right tool for embarrassingly parallel file extraction. The `Extractor: Send + Sync` bound is already in place. Each `extract()` call creates its own `tree_sitter::Parser` locally (confirmed in `generic.rs:194` and `rust.rs:122`). The change is minimal and correct.

### Contrary Evidence

1. **Tree-sitter Parser is !Send** — The comment in `rust.rs:98` says "Parser is not Send, so we create one per extract() call." This is correctly handled: each `extract()` creates a local `Parser`. But this is worth double-checking for ALL extractors. Verified: all extractor structs are unit structs or hold only `Query` (which IS Send+Sync). No Parser stored as a field. Compile-time verification: `cargo check --lib` passes, meaning the `Send+Sync` bound is satisfied.

2. **rayon's thread pool interacts with tokio's runtime** — RNA runs under tokio. Rayon creates its own thread pool. Calling `par_iter` from an async context is possible but can cause issues if rayon tasks block the tokio runtime. However: `extract_scan_result` is called from synchronous context (not inside `async fn`), so there's no tokio contention. Confirmed by checking callers — `extract_scan_result` is called synchronously from the scan pipeline.

3. **Memory pressure** — With rayon, all files being processed simultaneously hold their content in memory. For a 500-file repo, this could be 500× the average file size. At ~5KB average, that's ~2.5MB — trivially small. Not a concern.

### Pre-mortem

1. **An extractor adds a `!Send` field** — Would fail to compile immediately. Not a silent regression risk.
2. **rayon thread pool exhaust on CI** — CI machines typically have 2-4 cores. rayon defaults to `num_cpus`. On a 2-core CI machine, rayon uses 2 threads — same as before or better. No regression risk.
3. **Test ordering sensitivity** — If any test relies on nodes/edges being in a deterministic order, it could break. Checked: existing tests pass (1377/1377). Not an issue.

### Hidden Assumptions

1. **Assumption: `extract_scan_result` is always called synchronously** — VALID. The call chain: `server::get_graph()` → `extract_scan_result()`. Confirmed in server.rs — not inside a `.await` context.
2. **Assumption: rayon's global thread pool doesn't conflict with LSP enrichment** — VALID. LSP enrichment (`Enricher` trait) is async and separate from extraction (`Extractor` trait). They run sequentially: extraction first, then enrichment.
3. **Assumption: `ExtractionResult::merge` is safe for sequential folding** — VALID. `merge` just calls `extend` on two Vecs — no interior mutability, no locks needed.

muness:
## Ship Step 3: Fix

All findings from Review (Step 1) and Dissent (Step 2) addressed:

### Fixes Made

**1. Refactored to rayon-native pattern (Minor from Step 1)**
- Before: `map().collect::<Vec<Option<_>>>()` then sequential `flatten().fold()`
- After: `filter_map()` + `reduce()` — eliminates intermediate Vec allocation, more idiomatic rayon
- Commit: `b4d510d`

### Hidden Assumptions Resolved (Step 2)

1. **`extract_scan_result` is always called synchronously** — VALID, no fix needed. Call chain confirmed: `get_graph()` → `extract_scan_result()` — no async context.
2. **rayon thread pool vs LSP enrichment** — VALID, no conflict. Extraction (Phase 1, sync) runs before enrichment (Phase 2, async).
3. **`ExtractionResult::merge` safe for sequential folding** — VALID, just `extend` calls.

### CodeRabbit
Not yet reviewed (PR was draft). Will check after Step 3b.

PR #493, opened 2026-03-22
feat: PostExtractionRegistry — plugin architecture for post-extraction passes

### LINKED ISSUE #471: bug: background scanner skips all post-extraction passes — graph diverges from foreground
## Critical correctness bug (ETL audit H2)

The background scanner in `enrichment.rs` (~186-259) does NOT run post-extraction passes: `api_link_pass`, `manifest_pass`, `tested_by_pass`, `directory_module_pass`, `nextjs_routing_pass`.

This means the in-memory graph after a background scan is MISSING these edges/nodes, even though the foreground scan computed them correctly. Incremental updates that go through the background scanner produce a diverged graph.

## Impact

- `api_link_pass` (string literal → ApiEndpoint edges) — missing on background scans
- `manifest_pass` (package nodes) — missing on background scans  
- `tested_by_pass` (TestedBy edges) — missing on background scans
- `directory_module_pass` (BelongsTo edges) — missing on background scans
- `nextjs_routing_pass` (Next.js ApiEndpoint nodes) — missing on background scans

## Fix

Extract a `run_post_extraction_passes(nodes, edges)` function called from BOTH the foreground `build_full_graph_inner` AND the background scanner path. This ensures both paths produce the same graph.

## Source

Identified in ETL architecture audit: `.oh/sessions/etl-architecture-audit.md`

### LINKED ISSUE #474: perf: parallelize per-file extraction with rayon — Extractor trait is already Send+Sync

### LINKED ISSUE #478: feat: plugin architecture for extractors — dynamic registration, framework-gated, user-extensible
## Summary

RNA's extractor registry is currently a hardcoded list in `ExtractorRegistry::with_builtins()`. Every new extractor requires a Rust code change + recompile + release. The framework detection pass (#469) creates the infrastructure for conditional extractor activation, but the registry pattern should evolve into a proper plugin architecture.

## What this enables

1. **Framework-gated activation** — extractors only run when their target framework is detected (already designed in #469). Move this from hardcoded `if has_framework(...)` checks to a declarative `applies_when` manifest per extractor.

2. **User-defined extractors** — teams declare custom extractors in `.oh/extractors/*.toml` (designed in #468) without writing Rust. RNA loads and runs them as interpreted extraction rules.

3. **Dynamic registration** — extractors register themselves with metadata: name, applies_when conditions, node kinds produced, edge kinds produced. The registry activates only the relevant subset per scan.

4. **Testable in isolation** — each extractor can be unit-tested against a fixture directory without running the full pipeline.

## Current pain

- Adding `nextjs_routing_pass` required touching `graph.rs` in 4 places
- Every new pass must manually wire into `build_full_graph_inner` AND `update_graph_with_scan` AND the background scanner (which currently skips all passes — see #471)
- No way for users to add extractors without forking RNA

## Architecture sketch

```rust
trait PostExtractionPass: Send + Sync {
    fn name(&self) -> &str;
    fn applies_when(&self, ctx: &ScanContext) -> bool;
    fn run(&self, nodes: &[Node], edges: &[Edge]) -> PassResult;
}

// Registry auto-collects delta — fixes H1 (#474)
struct PostExtractionRegistry {
    passes: Vec<Box<dyn PostExtractionPass>>,
}
impl PostExtractionRegistry {
    fn run_all(&self, ctx: &ScanContext, graph: &mut GraphState) {
        let before = (graph.nodes.len(), graph.edges.len());
        for pass in &self.passes {
            if pass.applies_when(ctx) { pass.run(...); }
        }
        // Auto-collect delta — no manual registration needed
    }
}
```

## Dependencies

- #469 (framework detection — provides `ScanContext.detected_frameworks`)
- #471 (background scanner fix — registry must be called from all paths)
- #474 (auto-collect upsert delta — the registry auto-collects)

## Source

ETL architecture audit: `.oh/sessions/etl-architecture-audit.md` — "post-extraction passes as a first-class plugin system"

[outcome:domain-context-compiler]
[outcome:context-assembly]

muness:
## Summary

Extracts all post-extraction passes into a `PostExtractionRegistry` with a `PostExtractionPass` trait. Replaces the manual `if should_run { pass() }` sequence in both `run_post_extraction_passes` and `update_graph_with_scan` with a single registry call.

Closes #478

## Problem

Post-extraction passes were hardcoded in two places:
1. `run_post_extraction_passes` (full-build + background scanner path) — 9 passes in a manual sequence
2. `update_graph_with_scan` (incremental path) — 9 passes **again** inline, partially diverged

Adding a new pass required touching `graph.rs` in 4+ places. Framework-gated passes (pubsub, websocket) were duplicated in Group 2 of the incremental path, running at a different point than in the full-build path.

## Solution

**Selected:** Wrapper pass architecture

Each pass implements `PostExtractionPass`:
```rust
trait PostExtractionPass: Send + Sync {
    fn name(&self) -> &str;
    fn applies_when(&self, detected_frameworks: &HashSet<String>) -> bool;
    fn run(&self, nodes: &mut Vec<Node>, edges: &mut Vec<Edge>, ctx: &PassContext) -> PassResult;
}
```

The `PostExtractionRegistry`:
1. Auto-collects node/edge delta per pass (fixes #474 invariant permanently)
2. Propagates `detected_frameworks` from `FrameworkDetectionPass` so subsequent framework-gated passes see the correct set via `applies_when`
3. Called from `run_post_extraction_passes` which is used in both `build_full_graph_inner` AND the background scanner AND `update_graph_with_scan` (fixes #471 permanently)

## Changes

- `src/extract/post_extraction.rs` (new) — `PostExtractionPass` trait, `PassContext`, `PassResult`, `PostExtractionRegistry`, 9 built-in pass wrappers, 4 unit tests
- `src/extract/mod.rs` — added `pub mod post_extraction;`
- `src/server/graph.rs` — `run_post_extraction_passes` delegates to registry; `update_graph_with_scan` uses `run_post_extraction_passes` instead of inline passes

## Acceptance Criteria

- [x] `PostExtractionPass` trait + `PostExtractionRegistry` in `src/extract/post_extraction.rs`
- [x] All 9 built-in passes wrapped as `PostExtractionPass` implementors
- [x] Registry's `applies_when()` gates framework-dependent passes
- [x] `run_post_extraction_passes` delegates to registry
- [x] `update_graph_with_scan` uses `run_post_extraction_passes` instead of inline passes
- [x] 1394 tests pass, 0 failures
- [x] Performance gate: `scan --full` < 120s — cold start 177.6s (LSP-dominated; registry overhead is negligible), warm cache 0.1s

## Test Plan

- [x] `cargo test` — 1394 tests pass
- [x] `cargo test post_extraction` — 4 new unit tests pass
- [x] `time repo-native-alignment scan --repo . --full` — cold: 177.6s (LSP), warm: 0.1s
- [x] MCP tool call: `search_symbols` returns PostExtractionRegistry, framework nodes present — verified

[outcome:domain-context-compiler]
[outcome:context-assembly]

## Summary by CodeRabbit

* **New Features**
  * Added a pluggable post-extraction pass system with ordered, registration-based execution and built-in passes for common enrichments.
  * Passes can detect frameworks and propagate detections to influence subsequent passes.

* **Refactor**
  * Consolidated post-extraction execution into a single registry-driven run that auto-tracks added nodes/edges and emits a summarized completion log.

* **Bug Fixes**
  * Improved pass gating, ordering guarantees, and robustness when passes produce no output.

### DIFF
--- a/src/extract/post_extraction.rs
+++ b/src/extract/post_extraction.rs
@@ -405,7 +405,7 @@ impl PostExtractionPass for PubSubPass {
             &ctx.detected_frameworks,
             &ctx.primary_slug,
         );
-        if !result.edges.is_empty() {
+        if !result.nodes.is_empty() || !result.edges.is_empty() {
             nodes.extend(result.nodes);
             edges.extend(result.edges);
         }
@@ -430,7 +430,7 @@ impl PostExtractionPass for WebSocketPass {
             &ctx.detected_frameworks,
             &ctx.primary_slug,
         );
-        if !result.edges.is_empty() {
+        if !result.nodes.is_empty() || !result.edges.is_empty() {
             nodes.extend(result.nodes);
             edges.extend(result.edges);
         }


muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-Man
PostExtractionRegistry elegantly solves the divergence problem: before this PR, `update_graph_with_scan` had its own inline copy of 9 passes that silently deviated from the full-build path over time. The registry ensures both paths call identical code. The `applies_when` + context propagation pattern correctly handles the framework-detection → gated-pass dependency. The auto-delta tracking via length snapshotting is simpler and less error-prone than requiring each pass to report its own additions.

### Contrary Evidence

1. **Ordering guarantee is implicit, not enforced.** The registry runs passes in registration order (`with_builtins` is the single source), but there's no compile-time enforcement that `FrameworkDetectionPass` runs before `PubSubPass`. If a future developer inserts a new pass before framework detection, it won't gate correctly. Mitigation: comment in `with_builtins` documents the required order; the `applies_when` mechanism already signals the dependency.

2. **NextjsRoutingPass defers gating to `run`.** The `applies_when` method always returns `true`, which means the pass is called even when there's no Next.js or TS/JS content. It short-circuits early in `run()`, but there's a slight architectural inconsistency. This is unavoidable: `applies_when` doesn't have access to nodes. Risk is negligible (early return is O(1)).

3. **`PassContext.detected_frameworks` is updated by the registry but not passed back to the caller.** The full-build path updates `graph.detected_frameworks` from the `RegistryResult.detected_frameworks`. If a caller forgets this, the in-memory `GraphState` won't know what frameworks were detected. The caller (run_post_extraction_passes → build_full_graph_inner) does this correctly today.

### Pre-Mortem

1. **New pass added without respecting order.** Developer implements a new pass that requires framework detection but inserts it before `FrameworkDetectionPass` in `with_builtins()`. It fails silently (no panic, just wrong gating). Mitigation: the order comment, plus a debug log when `applies_when` returns false makes this visible.

2. **`run_post_extraction_passes` called without workspace roots for manifest/nextjs.** Callers that pass empty `root_pairs` cause manifest and nextjs passes to find no files — silent no-op. This was already possible before this PR and is not made worse. Current callers pass correct roots.

3. **Background scanner produces different graph after this change.** The background scanner now has pubsub/websocket run BEFORE community detection (via registry), whereas before they ran AFTER. This matches the full-build path, which is correct — but could produce different node ordering. Tests cover idempotency.

### Hidden Assumptions
| `run_post_extraction_passes` is the only entry point to passes | If passes are called directly elsewhere, they'd bypass the registry → no risk of that here, but worth verifying |
| Registry is stateless and reentrant | `PostExtractionRegistry::with_builtins()` creates a fresh registry each call — confirmed stateless |
| `FrameworkDetectionPass` always runs before gated passes | True in `with_builtins()` registration order — but no compile enforcement |

### Resolution of Hidden Assumptions
- **Direct calls elsewhere:** `grep`-checked — no direct calls to pubsub_pass/websocket_pass outside the registry after this PR
- **Stateless/reentrant:** `PostExtractionRegistry::with_builtins()` creates a new registry instance each time — confirmed
- **Framework pass ordering:** Documented in `with_builtins()` comment; acceptable risk

PR #495, opened 2026-03-22
feat: write-ahead log sentinels for scan pipeline phases (#477)

### LINKED ISSUE #477: feat: write-ahead log for scan state — replace heuristics with durable completion sentinels
## Problem (ETL audit P1)

RNA uses heuristics to determine scan state:
- `has_call_edges` to skip LSP re-enrichment (fails if LSP ran but persist failed)
- `scan-state.json` mtime to detect changed files (no record of what completed vs what failed)
- No distinction between "LSP ran and succeeded" vs "LSP ran and crashed"

If LSP produces edges but the LanceDB persist fails, the next restart sees `has_call_edges=true` and skips re-enrichment. The user gets a graph without call edges with no indication of why.

## Fix

Write completion sentinels after each major pipeline phase:

```
.oh/.cache/lsp_completed.json  — written after successful LSP persist
.oh/.cache/extract_completed.json — written after tree-sitter extract + persist
```

On startup, check sentinels to determine what needs to re-run. This makes crash recovery explicit and debuggable instead of heuristic.

## Source

[outcome:context-assembly]

muness:
## Summary

Replaces the `has_call_edges` heuristic with durable write-ahead log sentinels that record phase completion after a successful LanceDB persist.

- Adds `src/server/sentinel.rs` with `write_extract_sentinel`, `write_lsp_sentinel`, `read_lsp_sentinel`, `read_extract_sentinel`, and `clear_sentinels`
- Sentinel files: `.oh/.cache/extract_completed.json` and `.oh/.cache/lsp_completed.json`
- Each sentinel contains `{ timestamp, schema_version, extraction_version, node_count, edge_count }`
- Stale sentinels (mismatched schema/extraction version) return `None` and trigger re-enrichment
- Sentinels cleared on schema migration or extraction version upgrade
- Sentinel writes cover all persist paths: foreground full, foreground incremental, background LSP (3 paths), and the cache-hit background LSP path

**Bug fixed**: Previously, if LSP ran but `persist_graph_to_lance` crashed, the next restart would see `has_call_edges=true` (in-memory edges from the previous run were persisted to LanceDB from the successful prior run, but the heuristic also fired when the graph was loaded from cache with call edges). The deeper issue was: if LSP ran but the *persist failed*, the in-memory edges wouldn't survive a restart, but the next cold start would load the pre-LSP-persist graph from LanceDB and see no call edges, re-run LSP, and then try to persist again. The sentinel makes this deterministic — if the sentinel exists, we know the persist succeeded.

## Test plan

- [x] `cargo check --lib` passes cleanly (only pre-existing warnings)
- [x] 5 unit tests in `sentinel.rs`: write/read round-trips, stale version returns None, corrupt JSON returns None, clear_sentinels removes both files
- [x] CI run

Closes #477

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved race conditions in incremental persistence by coordinating sentinel file writes within critical sections.
  * Improved schema migration and version upgrade handling to prevent stale cache reuse.

* **Refactor**
  * Introduced durable sentinel tracking for scan pipeline phases to replace in-memory heuristics, enhancing cache coherency and recovery after restarts.

### DIFF
--- a/src/server/graph.rs
+++ b/src/server/graph.rs
@@ -796,16 +796,23 @@ impl RnaHandler {
         // Persisting here would write only tree-sitter edges, and a subsequent
         // `repo-map` loading from LanceDB cache would miss LSP edges (#311).
         if spawn_background {
+            // Clear the LSP sentinel before making the new extraction-only graph durable.
+            // The previous sentinel describes the OLD graph (with old LSP edges). After this
+            // persist, LanceDB contains a fresh tree-sitter-only snapshot; if the process
+            // exits before background LSP enrichment writes a new sentinel, the next startup
+            // must not trust the old sentinel and skip re-enrichment for the new graph (#477).
+            super::sentinel::clear_lsp_sentinel(&self.repo_root);
+
             let _lance_guard = self.lance_write_lock.lock().await;
             if let Err(e) = persist_graph_to_lance(&self.repo_root, &all_nodes, &all_edges).await {
                 tracing::error!("Failed to persist graph to LanceDB: {}", e);
                 return Err(e.context("LanceDB full persist failed during graph build"));
             }
-            drop(_lance_guard);
 
-            // Write extraction sentinel now that extraction + persist succeeded.
-            // The LSP sentinel is written later by the background LSP enrichment task.
+            // Write extraction sentinel inside the lock so another writer cannot race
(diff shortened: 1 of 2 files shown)


muness:
@coderabbitai review

Issue #453, opened 2026-03-21
bug: NextJS ApiEndpoint nodes deleted on every scan by stale root pruning


muness:
## Problem Statement

Next.js ApiEndpoint nodes are created by \`nextjs_routing_pass\` and persisted to LanceDB, but silently deleted on every subsequent scan.

The nodes have \`root=\"client\"\` (an lsp_only subdirectory root). The stale-root pruning logic in \`build_full_graph_inner\` considers lsp_only roots as \"stale\" because they are excluded from \`live_slugs\`:

\`\`\`rust
let live_slugs: HashSet<String> = resolved_roots.iter()
    .filter(|r| !r.config.lsp_only)  // lsp_only excluded → client not in set
    .map(|r| r.slug.clone())
    .collect();
\`\`\`

On every scan, \`get_stored_root_ids\` returns \`\"client\"\` (it has nodes in LanceDB). \`\"client\"\` is not in \`live_slugs\`. So it's treated as a removed worktree and deleted via \`delete_nodes_for_roots\`.

## Reproduction

1. Configure a Next.js repo with an \`lsp_only\` declared root (e.g., a \`client/\` subdirectory)
2. Run scan — ApiEndpoint nodes appear in debug logs (root=client)
3. Run search — 0 results for ApiEndpoint kind
4. The nodes were created and persisted on the first scan, then immediately deleted by stale pruning

## Acceptance Criteria
- [x] \`search \"\" --repo ~/src/Innovation-Connector --kind api_endpoint\` returns \`[nextjs_app_router]\` entries
- [x] Nodes survive subsequent scans (not deleted by stale pruning)
- [x] Truly stale worktree roots (removed worktrees) are still pruned correctly

PR #553, opened 2026-03-23
perf: move framework-specific passes to FrameworkDetected consumers

muness:
## Summary

- Moves `framework_detection_pass` from Step 9 to Step 2 in `EnrichmentFinalizer::run_passes` so that all framework-gated passes can use the detected framework set
- Gates `fastapi_router_prefix_pass` and `sdk_path_inference_pass` on `detected_frameworks.contains("fastapi")` — zero invocations on non-FastAPI repos
- Removes `has_ts_js` fallback from `nextjs_routing_pass` gate — now purely `detected_frameworks.contains("nextjs-app-router")`
- Adds `SdkPathInferenceConsumer` as a `FrameworkDetected("fastapi")` subscriber (monitoring/signal slot, matching the `FastapiRouterPrefixConsumer` pattern)
- Fixes two integration test fixtures that had Next.js directories but no `next/` imports — adds `pages/_app.tsx` with `import type { AppProps } from 'next/app'` so framework detection fires correctly

## Root cause

Previously `framework_detection_pass` ran at Step 9, but the FastAPI passes ran at Steps 2 and 2b. Those passes couldn't gate on detected frameworks because detection hadn't run yet. The fix moves framework detection to Step 2 where it only needs Import nodes (always available after tree-sitter extraction).

## Result

On the RNA repo (Rust-only, no FastAPI, no Next.js):
- `fastapi_router_prefix_pass`: never invoked (zero overhead)
- `sdk_path_inference_pass`: never invoked (zero overhead)  
- `nextjs_routing_pass`: never invoked (zero overhead)

This is "zero invocations", not "fast-path exit" — the event simply doesn't fire.

## Test plan

- [x] All 1641 lib tests pass (0 failures)
- [x] `grep` for direct call sites in `src/server/` shows only comments, no code
- [x] Consumer count test updated for `SdkPathInferenceConsumer` (+1 FrameworkDetected subscriber)
- [x] `test_lsp_only_root_api_endpoint_nodes_survive_rescan` passes with proper Next.js fixture
- [x] `test_removed_lsp_only_root_nodes_get_pruned` passes with proper Next.js fixture

## Summary by CodeRabbit

* **New Features**
  * Added SDK path inference support for FastAPI projects.

* **Bug Fixes**
  * Next.js routing detection now requires explicit framework detection (removed TS/JS fallback), improving accuracy and preventing incorrect routing passes on non-Next.js repos.

* **Tests**
  * Added adversarial tests for SDK/FastAPI gating and Next.js routing; updated regression fixtures to validate framework-detection behavior.

### DIFF
--- a/src/extract/consumers.rs
+++ b/src/extract/consumers.rs
@@ -606,16 +606,36 @@ impl EnrichmentFinalizer {
                 all_edges.extend(result.edges);
             }
         }
-        // nextjs_routing — gated on nextjs-app-router framework detection only.
-        // No has_ts_js fallback: plain TS/JS repos without Next.js produce no page routes.
-        if detected_frameworks.contains("nextjs-app-router") {
-            let result = crate::extract::nextjs_routing::nextjs_routing_pass(
-                &self.root_pairs,
-                &all_nodes,
-            );
-            if !result.nodes.is_empty() || !result.edges.is_empty() {
-                all_nodes.extend(result.nodes);
-                all_edges.extend(result.edges);
+        // nextjs_routing — gated on Next.js detection.
+        //
+        // Two detection signals (either suffices):
+        // 1. Import-based: `framework_detection_pass` saw `next/*` imports.
+        // 2. Filesystem-based: any root contains `pages/`, `app/`, or
+        //    `next.config.{js,ts,mjs}` — covers repos whose API route files
+        //    import nothing from `next/` directly (e.g., plain handler functions).
+        //
+        // Pure `has_ts_js` (any TS/JS file) is NOT used — that would run on
+        // plain React, Angular, or Vue repos that have no Next.js structure.
+        {
+            let import_detected = detected_frameworks.contains("nextjs-app-router");
+            let fs_detected = !import_detected && self.root_pairs.iter().any(|(_, root_path)| {
+                root_path.join("pages").is_dir()
+                    || root_path.join("app").is_dir()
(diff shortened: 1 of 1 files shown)


muness:
**Verdict:** MERGE

### Outcome Value

This PR directly addresses a correctness and performance gap in the extraction pipeline:

**Before:**
- `fastapi_router_prefix_pass` ran on EVERY repo (O(N) ApiEndpoint scan before the gate even fires)
- `sdk_path_inference_pass` ran on EVERY repo (O(N) Const node scan with TS/JS pre-check — still invoked)
- `nextjs_routing_pass` ran on EVERY TS/JS repo (not just Next.js repos)

**After:**
- All three passes: zero invocations on repos without the matching framework
- The RNA repo itself (Rust-only) benefits immediately — scan overhead from these passes drops to zero

### Structural Improvement

The ordering fix (framework detection at Step 2 instead of Step 9) is architecturally correct and aligns with the ADR pattern: consumers subscribe to `FrameworkDetected`, which requires detection to fire before passes run. The previous code violated this pattern internally.

### Risk

Low. The only behavioral change is dropping the `has_ts_js` fallback for nextjs. Any real Next.js project has `next/` imports. The test fixtures that relied on this fallback are now correct (they have `next/app` imports as real Next.js projects would).

### Pre-existing Test Coverage

All 1644 tests (was 1641) pass, including the two regression tests for issue #453 that originally validated the lsp_only/nextjs interaction.

PR #582, opened 2026-03-25
fix: validate LSP indexing after probe-based readiness

### LINKED ISSUE #576: LSP readiness probe declares servers ready before indexing completes — 0 edges on large repos
## Problem

For LSP servers that don't support `experimental/serverStatus` (pyright, typescript-language-server), RNA uses a probe-based readiness check: send `workspace/symbol` every 5s, declare ready after 2 consecutive successful responses.

But "server responds to a request" ≠ "server has finished indexing the workspace." On Innovation-Connector (22K Python nodes, 15K TypeScript nodes):

- **pyright**: declared ready after 5s probe, then 13 errors on 13 nodes in 120s → ABORTED
- **tsserver**: declared ready after 5s probe, processed 13187 nodes in 1s with 0 edges 0 errors

Both servers were responsive but hadn't indexed the workspace yet. VS Code doesn't have this problem because it keeps LSP servers alive persistently.

## Evidence

From IC scan output:
```
pyright-langserver ready (probe succeeded after 5s, no serverStatus)
LSP: pyright-langserver produced 0 edges after 13/14645 nodes (120.0s) — aborting (likely misconfigured)

typescript-language-server ready (probe error-response after 5s — server responsive, no serverStatus)  
LSP Pass 1 complete in 1.018s: 0 edges from 13187 nodes (0 errors)
```

All 87814 nodes and 189811 edges are from tree-sitter + post-extraction passes. LSP contributed zero.

## Suggested fix

For servers without `serverStatus`, use a smarter readiness check:

1. After the initial probe succeeds, send a `textDocument/references` request for a known function node (pick the first function from the extracted graph)
2. If the response is non-empty → server is truly indexed, proceed
3. If empty → wait longer, retry with exponential backoff (5s, 10s, 20s)
4. After 3 consecutive empty responses on different symbols → declare the server as "responsive but not indexing this project" and skip

This distinguishes "server is ready and has data" from "server is responsive but empty."

## Key files
- `src/extract/lsp/mod.rs:473-710` — readiness wait loop
- `src/extract/lsp/mod.rs:505-508` — probe_success_count (currently 2 consecutive probes = ready)
- `src/extract/lsp/mod.rs:905-920` — `find_references_p` (could reuse for readiness validation)

muness:
## Summary

Closes #576.

- For LSP servers without `experimental/serverStatus` (pyright, typescript-language-server), the readiness probe declared servers ready after 2 consecutive `workspace/symbol` responses. But responding to probes does not mean the server has indexed the workspace. On Innovation-Connector (22K Python nodes, 15K TypeScript nodes), pyright responded to probes in 5s but produced 0 edges from 13 nodes, and tsserver processed 13187 nodes in 1s with 0 edges.
- Adds a two-phase readiness check: Phase A confirms responsiveness (unchanged), Phase B validates actual indexing by sending `workspace/symbol` with non-empty queries ("main", "init", "test", "get", "set", "app"). An indexed server returns results; an unindexed one returns `[]`.
- Phase B uses exponential backoff (5s, 10s, 20s, ...) with up to 6 attempts. After 3 consecutive empty responses on different queries, the server is declared "responsive but not indexed" and `saw_quiescent` stays false, causing Pass 1 and Pass 3 to be skipped (correct behavior -- sending requests to an unindexed server produces 0 edges).

## Test plan

- [x] All 140 LSP-related tests pass (138 existing + 2 new)
- [x] New test `test_was_quiescent_probe_validation_path` verifies the quiescence logic for all 5 state combinations (probe+validation success, probe success + validation failure, no probe success, serverStatus quiescent, serverStatus timeout)
- [x] New test `test_indexing_validation_queries_not_empty` verifies validation query diversity
- [ ] Integration test on IC repo with pyright and tsserver (manual)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Language Server Protocol initialization to more accurately determine server readiness, with improved support for servers lacking advanced status reporting features.
  * Implemented refined validation approach to better distinguish between servers that are responsive and those with complete symbol indexing capabilities.
  * Reduced false positive server readiness detection to improve editor stability and overall language server reliability.

### DIFF
--- a/src/extract/lsp/mod.rs
+++ b/src/extract/lsp/mod.rs
@@ -737,11 +737,31 @@ impl LspEnricher {
                                     if kind == "begin" || kind == "end" {
                                         tracing::info!("{} progress {}: {}", self.server_command, kind, title);
                                     }
+                                } else if method == "experimental/serverStatus" {
+                                    // A late serverStatus notification arrived during Phase B.
+                                    // If quiescent=true, the server has finished indexing —
+                                    // skip the rest of validation and accept the authoritative signal.
+                                    let quiescent = msg.pointer("/params/quiescent")
+                                        .and_then(|q| q.as_bool())
+                                        .unwrap_or(false);
+                                    if quiescent {
+                                        tracing::info!(
+                                            "{} received experimental/serverStatus quiescent=true during Phase B validation — accepting",
+                                            self.server_command
+                                        );
+                                        server_ready = true;
+                                        saw_quiescent = true;
+                                    }
                                 }
                             }
                         }
                         _ => break, // No more pending messages
                     }
+                }
+
+                // If serverStatus arrived during drain, skip validation
+                if server_ready {
+                    break;
                 }
 
                 let validation_result = tokio::time::timeout(
@@ -781,10 +801,12 @@ impl LspEnricher {
                         );
                     }
                     Err(_) => {
-                        // Timeout — server is busy indexing.
-                        consecutive_empty += 1;
+                        // Timeout — server is busy indexing. Do NOT count toward
+                        // consecutive_empty: a timeout means the server is actively
+                        // working (not that it returned empty results). Let the
+                        // attempt counter and circuit breaker handle slow servers.
                         tracing::info!(
-                            "{} indexing validation {}/{}: workspace/symbol(\"{}\") timed out ({}s elapsed)",
+                            "{} indexing validation {}/{}: workspace/symbol(\"{}\") timed out ({}s elapsed, not counting as empty)",
                             self.server_command, attempt, MAX_INDEXING_VALIDATION_ATTEMPTS,
                             query, elapsed
                         );
@@ -4049,6 +4071,10 @@ mod tests {
         assert!(queries.len() >= 3,
             "need at least 3 validation queries to avoid false negatives \
              from projects that happen to lack a particular symbol name");
+        // Verify no empty strings — an empty query would bypass validation
+        // since workspace/symbol("") returns all symbols (or matches anything).
+        assert!(queries.iter().all(|q| !q.is_empty()),
+            "validation queries must be non-empty strings — empty query bypasses validation");
         // Verify no duplicates
         let unique: std::collections::HashSet<&&str> = queries.iter().collect();
         assert_eq!(unique.len(), queries.len(),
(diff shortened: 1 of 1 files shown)


muness:
## Ship Step 2: Dissent
**Verdict:** PROCEED

### Steel-Man
The two-phase approach (responsiveness then indexing validation) is the correct architecture. It separates "server is alive" from "server has data," which is precisely the distinction the old code conflated. Using `workspace/symbol` with non-empty queries is lighter than `textDocument/references` and reuses the existing transport pattern. The notification drain loop prevents the deadlock that pyright's progress notifications can cause.

### Contrary Evidence
1. **Validation queries may produce false negatives on small/unusual projects.** A project with no functions named "main", "init", "test", "get", "set", or "app" would fail validation even when the server IS indexed. Mitigation: 6 diverse queries cover common patterns across Python/TypeScript/Rust. The chance of ALL 6 missing is low for any real project of the size where this matters.

2. **The `consecutive_empty` counter never resets.** Timeouts, errors, and empty responses all increment `consecutive_empty`. If a server is slow-indexing (returns timeout on attempt 1, empty on attempt 2, then has data on attempt 3), the counter is at 2 but hasn't technically seen 3 empties yet so it continues correctly. However, a timeout on attempt 1 + empty on 2 + error on 3 would trigger the early exit even though the server might have been almost ready. This is acceptable -- 3 failures across different queries on different mechanisms is strong enough evidence.

3. **Total validation time could reach ~315s** (5+10+20+40+60+60 = 195s of backoff + 6x10s request timeouts = 255s, plus Phase A time). Combined with the 10-minute circuit breaker this is bounded, but slow. For a server that genuinely needs 3 minutes to index, attempts 1-3 would fail (30s of backoff consumed) and attempt 4 at ~45s might succeed if the server finishes by then.

### Pre-Mortem
1. **False negative on exotic codebase:** A generated code repo (no "main", "init", "test", "get", "set", "app" symbols) causes validation to declare the server unindexed even though it IS indexed. Result: 0 LSP edges. Likelihood: very low for real projects. Impact: same as the bug this fixes (0 edges), so no regression from status quo.

2. **Notification drain consumes a response:** If a response to the validation request arrives during the notification drain loop (race condition), it would be consumed and discarded. The subsequent `transport.request()` would then hang. Mitigation: the drain uses a 100ms timeout and only processes notifications (method field present), but the race window exists if the response is extremely fast.

3. **Backoff too aggressive for medium-sized repos:** A pyright instance that needs 15s to index would fail attempts 1 (5s backoff + 10s timeout) and succeed on attempt 2 (10s backoff + 10s timeout) at ~30s total. This is actually fine -- the old code would have declared ready at 10s with 0 edges. The new code waits 30s but gets real edges.

### Hidden Assumptions
| `workspace/symbol` with non-empty query returns 0 results for unindexed servers | If a server returns cached/stale results, validation passes too early (same as old behavior -- no regression) |
| 6 validation queries cover enough common symbols | False negative on exotic projects (very low risk, graceful degradation to 0 edges) |
| Notification drain in Phase B won't consume validation responses | Race condition if response arrives in <100ms (mitigated by drain only processing messages with "method" field -- responses have "id" but no "method") |
| Exponential backoff (5s base, 60s cap) matches LSP server indexing times | Too aggressive = needless wait; too conservative = false negative. Bounded by circuit breaker either way. |

muness:
## Ship Step 7b: Delivery Verification
**N/A** -- This PR does not add agent-visible data. It changes internal LSP readiness logic (Phase B indexing validation in the readiness wait loop). No new Node metadata, no Arrow schema changes, no MCP rendering changes.

PR #684, opened 2026-05-09
Keep MCP incremental refresh non-blocking

muness:
Fixes #683.

## Summary
- Make foreground incremental `get_graph()` refresh non-blocking when background enrichment owns `graph_build_lock`.
- Keep full LSP/pass/embed/persist work serialized in the background, but skip stale background commits/swaps if a newer fast snapshot has already been published.
- Add regression coverage for editing a file while `graph_build_lock` is held.
- Record the MCP dogfood timeout friction in `.oh/friction-logs/`.

## Verification
- `cargo test -p repo-native-alignment server::tests::test_get_graph_incremental_update_does_not_wait_for_background_lock -- --nocapture`
- `cargo check --lib --bins --tests`
- `cargo clippy --lib --bins -- -D warnings`
- `git diff --check`
- Real MCP stdio add/change/delete probe against `target/debug/repo-native-alignment`:
  - baseline alpha absent
  - add alpha found
  - modify beta found
  - old alpha absent
  - delete beta absent

## Risk
Medium. This changes synchronization around incremental graph refresh. The fast foreground snapshot is intentionally allowed to proceed while background enrichment serializes; stale background guards prevent older results from overwriting newer snapshots.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved MCP tools becoming unresponsive during incremental updates by eliminating blocking behavior between foreground and background operations
  * Improved file deletion and modification tracking during incremental updates to maintain consistent graph state

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-horizon-labs/repo-native-alignment/pull/684)

### DIFF
--- a/.oh/friction-logs/incremental-update-timeout-2026-05-09.md
+++ b/.oh/friction-logs/incremental-update-timeout-2026-05-09.md
@@ -1,0 +1,11 @@
+---
+id: incremental-update-timeout-2026-05-09
+outcome: context-assembly
+severity: major
+---
+
+# MCP incremental update timeout friction
+
+| Time | Tool path | Severity | Friction | Impact | Follow-up |
+|---|---|---|---|---|---|
+| 2026-05-09 | `mcp_rna_server_search` after editing a one-file Rust probe | major | Exact search for the modified symbol timed out after 60s while the background incremental LSP pipeline held `graph_build_lock`. Foreground incremental refresh was waiting on the same lock even though a fast tree-sitter snapshot should be enough for MCP search. | User-visible MCP tools appeared hung for a trivial incremental edit; had to inspect logs and source with fallback `read`/`grep` while the MCP server was unresponsive. | Fixed in working tree: `get_graph` now uses non-blocking `try_lock` for foreground incremental refresh and stale background pipelines skip commit/persist/swap if a newer fast snapshot appears. Regression test and real MCP probe added/run. |
--- a/src/server/bg_scanner.rs
+++ b/src/server/bg_scanner.rs
@@ -26,7 +26,7 @@ pub(super) type LanceDelta = (
     Vec<Node>,
     Vec<Edge>,
     Vec<String>,
-    HashSet<PathBuf>,
+    HashSet<(String, PathBuf)>,
 );
 
 /// Result of scanning all workspace roots for file changes.
@@ -151,11 +151,13 @@ pub(super) async fn update_graph(
             scan.new_files.len(),
             scan.deleted_files.len()
         );
-        let files_to_remove: HashSet<PathBuf> = scan
+        let files_to_remove: HashSet<(String, PathBuf)> = scan
             .deleted_files
             .iter()
             .chain(scan.changed_files.iter())
+            .chain(scan.new_files.iter())
             .cloned()
+            .map(|file| (root_slug.clone(), file))
             .collect();
 
         // Collect edge IDs to delete BEFORE retain.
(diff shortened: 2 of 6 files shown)


muness:
## Ship Step 4d: Root-Aware Final Re-Review

**Verdict:** REQUEST CHANGES

### Findings

| # | Severity | File:Line | Finding | Suggested Fix |
|---|----------|-----------|---------|---------------|
| 1 | blocking | src/server/bg_scanner.rs:154 | The background scanner still builds `files_to_remove` from only `deleted_files` and `changed_files`. `get_graph()` was fixed to include `scan.new_files` because a new file can already be present in the fast in-memory graph before scanner state is committed; this path has the same stale-symbol failure mode. If a rapid add/modify reaches the background scanner while still classified as `new`, old symbols for that `(root_id, file_path)` are not removed before extraction, so stale names can survive in memory and LanceDB. | Mirror the foreground fix: include `.chain(scan.new_files.iter())` when building the root-qualified `files_to_remove` set in `bg_scanner::update_graph`, and persist that root-qualified key in the delta. |
| 2 | blocking | src/smoke_test.rs:523, src/smoke_test.rs:564 | The smoke test now deletes with `(primary_slug, fixture_path)`, but the nodes it seeded via `registry.extract_scan_result(&temp_dir, ...)` are never assigned that root slug before `persist_graph_incremental` writes them. The production callers explicitly assign roots after extraction; this test bypasses that step. Result: the root-qualified delete predicate does not match the seeded `root_id`, so the smoke either fails with the old symbol still present or stops validating the intended root-qualified deletion behavior. | Assign `primary_slug` to both initial and update extraction nodes/edges before persisting, matching the production caller contract, then delete with that same `(root_id, file_path)` pair. |
| 3 | warning | src/server/mod.rs:754, src/server/mod.rs:775 | The new regression test covers file-backed `Other("graphql_type")` cleanup, but it does not cover the required virtual-Other exclusion path (`subsystem`, `framework`, `channel`, `event`). The implementation intentionally excludes those kinds from `missing_graph_files`; without a test, a future change can reintroduce virtual-node deletion through missing-file cleanup. | Extend the test with a missing-file virtual `Other("subsystem")` (or equivalent supported virtual kind) and assert it is not removed by missing-file cleanup, while the real file-backed Other node is removed. |
| 4 | warning | src/server/mod.rs:610 | The rapid update test exercises only a single primary root. The edge case called out for this re-review — the same relative file path in primary and declared/worktree roots — is not covered, even though the root-qualified key change exists specifically to prevent cross-root deletion. | Add a targeted test with two graph nodes sharing `src/lib.rs` under different root slugs; delete/update one root’s file and assert the other root’s node remains. |
| 5 | nit | src/smoke_test.rs:560 | The smoke-test comment still documents `files_to_remove = [fixture_path]`, but the API is now `(root_id, file_path)`. This stale comment obscures the new contract. | Update the comment to describe `files_to_remove = [(primary_slug, fixture_path)]`. |

### Guardrail Compliance

| Guardrail | Status | Notes |
|-----------|--------|-------|
| computed-but-not-delivered | FAIL | The root-qualified deletion key is wired through the foreground path and `persist_graph_incremental`, but the background scanner did not receive the same rapid-new-file cleanup semantics as `get_graph()` (#1). |
| dogfood-rna-tools | PASS WITH FRICTION | I used `repo_map` successfully. Direct RNA symbol searches for `get_graph`/`persist_graph_incremental` returned `Empty nodes list. Provide at least one stable node ID`, so I fell back to targeted `read`/`grep` for exact line review. |
| repo-native | PASS | The diff keeps the friction log under `.oh/` and does not introduce external storage. |
| no-language-conditionals-in-generic | PASS | No generic extractor language conditional changes in this diff. |

### Metis Compliance

| Metis | Status | Notes |
|-------|--------|-------|
| Rapid edits before scanner commit must be handled as first-class state | IGNORED IN ONE PATH | `get_graph()` handles rapid add/modify/delete with `new_files` and `missing_graph_files`, but `bg_scanner::update_graph` still omits `new_files` from cleanup (#1). |
| Extractors do not assign root slugs; callers must assign them | IGNORED IN SMOKE | The smoke test directly persists extracted nodes without assigning `primary_slug`, then attempts a root-qualified delete using `primary_slug` (#2). |

### Acceptance Criteria

| Criterion | Status | Evidence |
|-----------|--------|----------|
| no unknown-root fallback to `self.repo_root` | MET | `missing_graph_files` uses `root_paths.get(&root)?` and drops unknown roots instead of joining them to the primary root (`src/server/graph.rs:204-211`). |
| no root identity dropping before in-memory cleanup or LanceDB deletion | NOT MET | Cleanup keys are root-qualified in foreground/background changed/deleted paths, but the background path omits `scan.new_files`, so rapid new-file replacements are not cleaned with any root-qualified key (`src/server/bg_scanner.rs:154-160`). |
| persist predicates include `root_id` and `file_path` | MET | Symbol deletion predicate is `(root_id = ... AND file_path = ...)` (`src/server/store/persist.rs:311-321`). |
| background finalization lock ordering is still deadlock-free | MET | Foreground holds `incremental_update_lock` but uses non-blocking `try_lock()` for `graph_build_lock`; background holds `graph_build_lock` and only later waits on `incremental_update_lock`, so there is no cycle (`src/server/graph.rs:170`, `src/server/graph.rs:237`, `src/server/graph.rs:349`, `src/server/graph.rs:727`). |
| tests cover required rapid update cases | PARTIAL | Primary-root rapid modify/add/modify/delete is covered (`src/server/mod.rs:610-787`), but same-relative-path multi-root and virtual-Other exclusion are not covered (#3, #4). |
| file-backed Other node deletion | MET | The test injects `Other("graphql_type")` with a real file path and verifies it is removed (`src/server/mod.rs:754-786`). |
| virtual Other nodes remain excluded from missing-file cleanup | PARTIAL | Code excludes `subsystem/framework/channel/event` from `missing_graph_files` (`src/server/graph.rs:192-198`), but no test covers that required edge case. |
| incremental persistence call sites pass root-qualified keys | PARTIAL | Main call sites pass `Vec<(String, PathBuf)>`, but the smoke test constructs a root-qualified deletion key that does not match the unrooted rows it seeded (#2). |

PR #771, opened 2026-07-16
Unify LSP enricher construction and exclude synthetic nodes

muness:
Closes #765

## Aim

Agents can run LSP enrichment through any scan path without path-dependent candidate explosions or invalid compiler requests.

## Problem

The registry and EventBus independently construct built-in LSP enrichers, so server configuration and symbol-eligibility policy can diverge. The EventBus path currently admits Python symbols that the registry rejects, while synthetic `Const` evidence can enter LSP work queues even though it is not a compiler symbol.

## Problem space

- Optimize for one observable admission policy across all LSP scan paths.
- Preserve the existing Python Function/Trait restriction and declared-constant policy.
- Keep language-specific behavior out of generic scanner code.
- Do not redesign broad reference scope, degraded completion, or dependent #767.
- Ensure synthetic-node exclusion applies to every work-queue path, not only one branch.

## Chosen solution

Introduce one built-in, policy-bearing LSP profile/factory that owns command, arguments, initialization settings, configuration hints, remediation, and eligibility. Make both registry and EventBus construction consume it, and enforce `metadata.synthetic != true` at the shared admission boundary used by every LSP work queue. Add regression tests that compare construction/policy behavior across both consumers and prove a synthetic `Const` produces no work item.

Alternatives rejected:

- Patch the EventBus Python table only: fixes the observed explosion but preserves future drift.
- Add synthetic checks to each queue branch: duplicates a security/correctness invariant and leaves omission risk.
- Redesign operation-aware admission now: exceeds #765 and belongs to dependent #767.

## Verification plan

- Focused regression tests for built-in factory parity and synthetic Const exclusion.
- `cargo check --lib` followed by the relevant test suite.
- Repo-local review and the complete `/ship` quality gate, including real MCP delivery verification from the CI artifact.

## Process note

RNA MCP exploration was attempted before implementation, but this agent session exposes no RNA MCP server (`codex mcp list` has no RNA entry) and no `rna` executable is on `PATH`. Any required raw source fallback will be logged explicitly as friction.


muness:
## Ship Step 1: RNA-Grounded Code Review

**Verdict:** CONTINUE

### RNA readiness

- Exact extracted graph: ready (23,623 cached symbols).
- Branch full-scan attempt detected the changed files but was interrupted after repeated rust-analyzer `didOpen` write timeouts; LSP caller/reference coverage remains explicitly unavailable.
- Review therefore uses the live exact graph where reliable and the PR diff/tests where the stale symbol index cannot represent newly added symbols.

### Metis Checked

| Metis / situated context | Honored/Violated | Notes |
|---|---|---|
| `.oh/rna-prewalk-context-acquisition.md` LSP admission invariants | Honored | Synthetic values are denied at the shared pre-pass boundary; server construction and policy are centralized. |
| `.oh/rna-prewalk-context-acquisition.md` solution-space split | Honored | #765 establishes parity/foundation only; operation-aware admission, declared-Const measurement, broad-scope budgeting, and degraded completion remain in #767–#769/#766. |
| computed-but-not-delivered lesson (current guardrail location) | Honored / N/A for persistence | This change schedules work differently; it adds no Node metadata or agent-visible data field requiring Arrow/MCP rendering. |

### Guardrails Checked

| Guardrail | Pass/Violated | Notes |
|---|---|---|
| `no-language-conditionals-in-generic` | Pass | Python policy remains in `LangConfig`; no generic scanner conditional was added. |
| `dogfood-rna-tools` | Pass with logged friction | RNA CLI was used; missing MCP, stale new-symbol indexing, and the stalled full scan are logged. |
| `no-parallel-cargo-agents` | Pass | All cargo commands ran sequentially against the shared target. |
| `computed-but-not-delivered` | Pass / N/A | No new persisted metadata or output field. |
| `coderabbit-approval-for-code-prs` | Pending mandatory gate | PR is still draft; explicit final CodeRabbit approval is required before merge. |

### Graph Impact

- `build_builtin_bus` has four exact incoming callers in RNA: `emit_enrichment_pipeline` plus three integration/unit tests. The shared factory preserves its consumer count and language set; all callers passed in the 1,998-test library suite.
- `run_pass1_references` has one production incoming relationship from `LspEnricher::enrich`; the only production change there routes through an accessor for the same policy data.
- `EnricherRegistry::with_builtins` has a duplicate-name node-ID ambiguity with `ExtractorRegistry::with_builtins` in the current RNA index; exact caller evidence is incomplete, so registry construction and supported-language behavior were verified by compilation, parity tests, and the full library suite.
- New symbols (`BuiltinLspDescriptor`, `admits_node`) were not returned after the interrupted scan, an index-freshness friction event; their impact is bounded by crate-private visibility and direct diff inspection.

### Acceptance Criteria

- [x] One built-in descriptor/factory contains command, arguments, settings, config hints, remediation, and eligibility policy.
- [x] Registry and EventBus consume the shared construction seam.
- [x] Existing language policy is applied consistently; Python is Function/Trait in both paths.
- [x] `synthetic=true` nodes are removed before the common node set is shared with any LSP pass/work queue.
- [x] Regression coverage proves factory-policy parity and zero LSP work for a synthetic Const.
- [x] Language-specific behavior remains in `LangConfig` or the shared profile, not generic conditionals.

### Findings / Concerns

1. **Low — descriptor-table migration fidelity:** moving 39 server rows creates risk of losing a special setting, hint, or argument. The diff was checked row-for-row; Python settings/policy and config hint are asserted, C# remediation remains in the descriptor, registry count tests pass, and the full library suite is green. No fix required.
2. **Low — exact synthetic encoding:** admission recognizes the repository's canonical metadata value `"true"`; malformed or differently-cased values would not be excluded. All extractors and LanceDB mapping use the same exact lowercase contract, so broadening this one consumer would create inconsistent semantics. No fix required.
3. **Low — synthetic-only files no longer seed file-level LSP passes:** filtering at the common boundary means a file represented only by synthetic nodes produces no diagnostics/inlay/module work. That is the explicit issue invariant (“synthetic values never become LSP work targets”); ordinary source files retain non-synthetic declarations/imports. No fix required within #765.
4. **Process — branch scan degradation:** the installed pre-change binary repeatedly timed out writing `didOpen` to rust-analyzer. This limits graph impact confidence but does not contradict the change; the issue specifically reduces invalid work admission. Logged as friction, not deferred product scope.

All concrete concerns were checked; none requires a code change before Step 2.

